PR #279, opened 2025-08-18
Conn state

eivindj-nordic:


MirkoCovizzi (commented):
One note, I think the folder should be also named `ble_conn_state`.

lib/ble_conn_state/ble_conn_state.c
```diff
@@ -0,0 +1,424 @@
...
+int ble_conn_state_user_flag_acquire(void)
+{
+	for (int i = 0; i < CONFIG_BLE_CONN_STATE_USER_FLAG_COUNT; i++) {
+		if (atomic_test_and_set_bit(&bcs.acquired_flags, i) == false) {
+			/* Bit was not set before */
+			return i;
+		}
+	}
+
+	return -ENOSPC;
```
MirkoCovizzi, line 334: This is going against the recent discussions we've had about error codes direction for BLE related components. Is there some other way to handle this? Since Peer Manager depends on this component, I would be careful with changing return values or all the changes will have to be propagated to the Peer Manager and will have to be debugged at runtime due to possible subtle bugs (since `uint32_t` vs `int` compiles ok).

eivindj-nordic: Discussed online and agreed. Replaced with BLE_CONN_STATE_USER_FLAG_INVALID.

lib/ble_conn_state/ble_conn_state.c
```diff
@@ -0,0 +1,424 @@
...
+#include <ble.h>
+#include <ble_gap.h>
+#include <nrf_sdh_ble.h>
+#include <zephyr/sys/atomic.h>
+#include <zephyr/logging/log.h>
+#include <errno.h>
+
+LOG_MODULE_REGISTER(ble_bms, CONFIG_BLE_CONN_STATE_LOG_LEVEL);
+
+#define BLE_CONN_STATE_DEFAULT_FLAG_COLLECTION_COUNT 6
```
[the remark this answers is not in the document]
eivindj-nordic: No, it is the number of flags in the struct below before the user flags.

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -28,8 +28,11 @@ SoftDevice Handler
 ==================
 
 * Added the :c:func:`nrf_sdh_ble_conn_handle_get` function.
-* Updated the system initialization to initialize on application level.
-* Updated the entropy source to use the CRACEN true random number generator instead of the CRACEN pseudorandom number generator.
+
+* Updated:
+
+   * The system initialization to initialize on application level.
+   * The entropy source to use the CRACEN true random number generator instead of the CRACEN pseudorandom number generator.
```
anhmolt, line 35: Can you put this in a separate commit?

PR #307, opened 2025-08-22
samples: bluetooth: ble_pwr_profiling: Turn off led as default and update adv data

eivindj-nordic:
When on, LED0 draws 1.8uA due to the 1M pulldown. This increases the current by 50% in CPU idle state from ~4uA to ~5.8uA. This commit turns the LED off by default to match the behaviour in NCS.

* Change default advertisement name to Nordic_Power_Profiling
* Remove service from advertising data.

This aligns the advertising data with NCS samples, which gives a better
comparison between the two.

### DIFF
--- a/samples/bluetooth/ble_pwr_profiling/README.rst
+++ b/samples/bluetooth/ble_pwr_profiling/README.rst
@@ -122,6 +122,11 @@ LED 0:
    This causes a delay between the SoC wake up and button press processing.
    If you want to start advertising on system start, you must keep pressing the button until you see a log message confirming the advertising start on the terminal.
 
+.. note::
+   The LED is disabled by default.
+   You can enable the LED by setting the :kconfig:option:`CONFIG_BLE_PWR_PROFILING_LED` Kconfig option.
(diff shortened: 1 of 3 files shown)


samples/bluetooth/ble_pwr_profiling/README.rst
```diff
@@ -122,6 +122,9 @@ LED 0:
    This causes a delay between the SoC wake up and button press processing.
    If you want to start advertising on system start, you must keep pressing the button until you see a log message confirming the advertising start on the terminal.
 
+.. note::
+   To observe the LED the :kconfig:option:`CONFIG_BLE_PWR_PROFILING_LED` Kconfig option must be set.
```
anhmolt, line 126: Can add a line here telling that the current consumption will increase if the LED is turned on.

anhmolt (approved):

samples/bluetooth/ble_pwr_profiling/README.rst
MirkoCovizzi: I feel like this sentence is missing something such as `To observe the LED being turned on...`

MirkoCovizzi (approved):

PR #310, opened 2025-08-22
drivers: add empty GPIO driver

eivindj-nordic:
Add empty GPIO driver. This removes a compile warning for MCUboot.


MirkoCovizzi (approved):

nordicjm (changes requested):
let's not add hacks

PR #315, opened 2025-08-25
samples: peripheral: bm_zms: use subpartition and move configuration

eivindj-nordic:
* Use subpartition for the ZMS instance as this is what we expect the users to do.
* Move sector size and count configuration to Kconfig.

### DIFF
--- a/samples/peripherals/bm_zms/src/main.c
+++ b/samples/peripherals/bm_zms/src/main.c
@@ -7,8 +7,11 @@
 #include <zephyr/logging/log.h>
 #include <zephyr/logging/log_ctrl.h>
 
-#define STORAGE_NODE		DT_NODELABEL(storage_partition)
-#define BM_ZMS_PARTITION_OFFSET	DT_REG_ADDR(STORAGE_NODE)
+#define STORAGE_NODE DT_NODELABEL(storage0_partition)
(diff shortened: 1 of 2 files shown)


eivindj-nordic:
> I think this requires a few more constraints. E.g. should the Kconfigs have ranges?

Constraints given by the build_assert. User should be able to set whatever they want to test, as long as it is within the partition size.

eivindj-nordic:
> I understand, but there are a few more constraints than that. For example the size of the sector needs to satisfy some requirements. However, I have not tested if ZMS itself will prevent mounting in that case and return an error.

@MirkoCovizzi If there are other internal constraints for the sector size in ZMS, that should be handled accordingly by the ZMS library I think.

samples/peripherals/bm_zms/src/main.c
```diff
@@ -163,10 +163,10 @@ int main(void)
 	}
 
 	fs.offset = BM_ZMS_PARTITION_OFFSET;
-	fs.sector_size = 1024U; /* 1kB sector size. */
-	fs.sector_count = 3U;
+	fs.sector_size = CONFIG_BM_ZMS_SECTOR_SIZE;
+	fs.sector_count = CONFIG_BM_ZMS_SECTORS_COUNT;
```
nordicjm, line 167: can't this just be calculated?

PR #324, opened 2025-08-26
sysbuild: Pass SoftDevice Kconfig selection over

nordicjm:

PR #326, opened 2025-08-26
Program SoftDevice with application image (none MCUboot board variants)

anhmolt:
Add a new Kconfig option `CONFIG_PROGRAM_SOFTDEVICE_WITH_APP`. When enabled, the SoftDevice hex will be programmed together with the application image. This only applies to the none MCUboot board variants.

depends on https://github.com/nrfconnect/sdk-nrf-bm/pull/324


cmake/merge_app_softdevice.cmake
```diff
@@ -0,0 +1,33 @@
...
+    ${CONFIG_SOFTDEVICE_FILE}
+  )
+
+  set_target_properties(runners_yaml_props_target PROPERTIES "hex_file"
+    "${ZEPHYR_BINARY_DIR}/${KERNEL_NAME}_softdevice_merged.hex"
+  )
+  set(BYPRODUCT_KERNEL_APP_SD_HEX_NAME
+    "${ZEPHYR_BINARY_DIR}/${KERNEL_NAME}_softdevice_merged.hex"
+    CACHE FILEPATH "Application with SoftDevice hex file" FORCE
+  )
```
nordicjm, line 31: we don't need another merged file like this no

nordicjm (changes requested):
do this properly with a dedicated image, use https://github.com/zephyrproject-rtos/zephyr/pull/94796 to see how to create and flash a fake image which isn't actually a zephyr project

eivindj-nordic:
Replaced by #450

PR #330, opened 2025-08-27
Unittest fixes

MechanicV1:


eivindj-nordic (approved):

MirkoCovizzi (commented):
There was a reason why the tests for `bm_storage` were placed there. It is because we should have in that folder also the 2 tests for the 2 backends. I personally see little reason to change it now here and then changing it back again once we add those tests.

eivindj-nordic:
Will keep storage test location as is and remove timer test instead (#335).

PR #331, opened 2025-08-27
doc: update doxygen for S115

b-gent:
Doxygen adjustments for publishing S115 API doc.


knutel-nordic (approved):
There are some issues with literal HTML tags in a few places, but we agreed it will be fixed in a later PR.

PR #332, opened 2025-08-27
doc: add a topic to dfu doc

b-gent:
Add more info on single bank dfu with diagrams.


b-gent:
@nordicjm  can you have a look ?

PR #341, opened 2025-08-28
doc: doc update after 0.8.0

b-gent:
Update version and bring back migration guide.


MirkoCovizzi (approved):
This should go after we update the version number to `0,8.99`

PR #353, opened 2025-09-04
lib: bm_zms: reword `bm_zms_evt_t` member names

MirkoCovizzi:
* Rewords the event ID `id` to `evt_id`

* Rewords the entry ID `ate_id` to `id`

This aligns the `id` parameters for write/delete
operations with the name of the event struct member for the corresponding ID reported.

### DIFF
--- a/include/bm_zms.h
+++ b/include/bm_zms.h
@@ -41,9 +41,11 @@ typedef enum {
 
 /**@brief A BM_ZMS event. */
 typedef struct {
-	bm_zms_evt_id_t id; /* The event ID. See @ref bm_zms_evt_id_t. */
-	uint32_t result;    /* The result of the operation related to this event. */
-	uint32_t ate_id;    /* The ATE id in case of a write operation. */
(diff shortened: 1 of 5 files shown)


include/bm_zms.h
```diff
@@ -40,9 +40,11 @@ typedef enum {
 
 /**@brief A BM_ZMS event. */
 typedef struct {
-	bm_zms_evt_id_t id; /* The event ID. See @ref bm_zms_evt_id_t. */
-	uint32_t result;    /* The result of the operation related to this event. */
-	uint32_t ate_id;    /* The ATE id in case of a write operation. */
+	bm_zms_evt_id_t evt_id; /* The event ID. See @ref bm_zms_evt_id_t. */
```
b-gent, line 43: These are not doxygen comments but standard comments, hence doc for these members is not rendered.
See https://www.doxygen.nl/manual/docblocks.html#memberdoc

This is also why they do not link from the changelog.

I can see this occurring in some more places across this file and I don't mean that we need to fix this as part of this PR, just pointing it out.

PR #359, opened 2025-09-16
ci: Use fixed version of toolchain bundle

jangalda-nsc:
Use fixed version of toolchain bundle
in src.tar.gz package

Please remember to update bundle version when needed
It should also match toolchain version used in twister tests
https://github.com/nrfconnect/sdk-nrf-bm/blob/main/.github/workflows/twister.yml#L16

It is possible to extract bundle version to separate file, but it requires more changes in actions and is out of scope of my PR


eivindj-nordic (approved):
Thanks!

MirkoCovizzi (approved):

PR #366, opened 2025-09-18
Lesc export key

eivindj-nordic:


lib/peer_manager/modules/nrf_ble_lesc.c
```diff
@@ -144,6 +151,24 @@ uint32_t nrf_ble_lesc_keypair_generate(void)
 		LOG_ERR("psa_export_public_key() returned status %d", status);
 		return NRF_ERROR_INTERNAL;
 	}
+
+#if defined(CONFIG_PM_LESC_PRIVATE_KEY_EXPORT)
+	uint8_t priv_key[ECC_PRIV_KEY_EXPORT_SIZE];
```
MirkoCovizzi, line 156: I would also put a big log disclaimer. Putting it in the Kconfig description is not visible enough.

tejlmand (approved):

lib/peer_manager/modules/nrf_ble_lesc.c
```diff
@@ -144,6 +151,26 @@ uint32_t nrf_ble_lesc_keypair_generate(void)
...
+	size_t priv_key_len = 0;
+
+	LOG_WRN("CONFIG_PM_LESC_PRIVATE_KEY_EXPORT is not to be used in production!");
```
MirkoCovizzi, line 159: I would use the most available log level to make sure it is shown.

PR #391, opened 2025-09-24
sysbuild: bootconf lock

michalek-no:
Add bootloader immutability through RRAMC's region no.3 Applies before first boot.


cmake/sysbuild/bootconf.cmake
```diff
@@ -0,0 +1,30 @@
+#
+# Copyright (c) 2025 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+function(setup_bootconf_data)
+    add_custom_target(bootconf_target
```
nordicjm, line 8: cmake indent is 2 spaces, also move file up a folder as this does not run in sysbuild context

cmake/sysbuild/bootconf.cmake
```diff
@@ -0,0 +1,30 @@
...
+      ALL
+      DEPENDS ${CMAKE_BINARY_DIR}/bootconf.hex
+    )
+
+  dt_nodelabel(boot_partition_node_full_path NODELABEL "boot_partition")
```
nordicjm, line 13:
```suggestion
  dt_nodelabel(boot_partition_node_full_path NODELABEL "boot_partition" REQUIRED)
```

PR #393, opened 2025-09-24
subsys: softdevice_handler: remove the limits of link count on s115

laxiLang:
There is no need setting the upper limit of total link count (CONFIG_NRF_SDH_BLE_TOTAL_LINK_COUNT) to 1 for s115.


nordicjm:
How many connections does the s115 softdevice support?

eivindj-nordic:
Addressed in #483.

PR #394, opened 2025-09-24
softdevice: imply NRF_SDH

lemrey:
Imply that selecting `SOFTDEVICE` enables `NRF_SDH`.
That's a safe assumption when `CONFIG_SOFTDEVICE` is enabled in a project.


nordicjm (approved):

MirkoCovizzi (commented):
What about those libraries or components that right now are expressing direct dependency on both `CONFIG_SOFTDEVICE` and `CONFIG_NRF_SDH`?

lemrey:
Good point, I should double check and see that all components depending on `NRF_SDH` via an observer express that dependency via Kconfig.

eivindj-nordic (approved):
@lemrey Please add changelog entry.

anhmolt (approved):

PR #399, opened 2025-09-25
Build repeated pairing attempt protection variants

MirkoCovizzi:


samples/bluetooth/ble_hids_keyboard/sample.yaml
```diff
@@ -19,3 +19,23 @@ tests:
...
+    build_only: true
+    integration_platforms:
+      - bm_nrf54l15dk/nrf54l05/cpuapp/s115_softdevice
+      - bm_nrf54l15dk/nrf54l10/cpuapp/s115_softdevice
+      - bm_nrf54l15dk/nrf54l15/cpuapp/s115_softdevice
+      - bm_nrf54l15dk/nrf54l05/cpuapp/s115_softdevice/mcuboot
+      - bm_nrf54l15dk/nrf54l10/cpuapp/s115_softdevice/mcuboot
+      - bm_nrf54l15dk/nrf54l15/cpuapp/s115_softdevice/mcuboot
+    platform_allow:
+      - bm_nrf54l15dk/nrf54l05/cpuapp/s115_softdevice
```
eivindj-nordic, line 33: Do we need to build this for all platforms?

MirkoCovizzi: That depends. We don't have a consistent policy on how to go about this.

eivindj-nordic: Yes, we need to decide. Building this configuration six times per sample seems a bit of a waste as nothing in the board itself should affect it, other than the memory size.

eivindj-nordic (approved):

eivindj-nordic:
@MirkoCovizzi Been thinking a bit and discussing with @anhmolt. I think this should be verified in a test instead. We cannot use the samples to test Kconfigs. Kconfig fragments that we are expecting the users to add to the configuration is a different case.

PR #412, opened 2025-10-03
sysbuild: load CRACEN microcode once

michalek-no:
select which image initializes CRACEN


michalek-no:
> @michalek-no What is the status on this?

now it should be good to go

PR #416, opened 2025-10-08
doc: nrf-bm: lib: bluetooth: add documentation for queued writes

eivindj-nordic:
Add documentation for BLE queued writes library.


doc/nrf-bm/doxygen/nrf-bm.doxyfile
```diff
@@ -2401,6 +2401,7 @@ INCLUDE_FILE_PATTERNS  =
 # This tag requires that the tag ENABLE_PREPROCESSING is set to YES.
 
 PREDEFINED             = __DOXYGEN__ \
+                         CONFIG_BLE_QWR_MAX_ATTR=1 \
```
b-gent: Adding this because the register function was surrounded by an IF which prevented it from being picked up by doxygen.

b-gent (approved):

PR #438, opened 2025-10-22
Peer Manager typedef removal and some refactoring

anhmolt:
Typedef removal from PM API and internal. Plus some refactoring.

- [x]  squash the typedef removal commits to one or two commits when review is done.


b-gent (approved):

eivindj-nordic (approved):
Left some nits, would at least be good to align the struct name (adding _evt as mentioned above).

lemrey (approved):
Thank you!

PR #449, opened 2025-10-28
Peer Manager: fixes for central

anhmolt:
Fixes for using Peer Manager with S145 SoftDevice.

### DIFF
--- a/lib/bluetooth/peer_manager/modules/security_dispatcher.c
+++ b/lib/bluetooth/peer_manager/modules/security_dispatcher.c
@@ -238,6 +238,25 @@ static void sec_proc_start(uint16_t conn_handle, bool success, enum pm_conn_sec_
 	}
 }
 
+/**
+ * @brief Function for initiating pairing as a central, or all security as a periheral.
+ *
+ * See @ref smd_link_secure and @ref sd_ble_gap_authenticate for more information.
+ */
+static uint32_t link_secure_authenticate(uint16_t conn_handle, ble_gap_sec_params_t *sec_params)
+{
+	uint32_t err_code = sd_ble_gap_authenticate(conn_handle, sec_params);
+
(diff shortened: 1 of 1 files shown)


nordicjm (approved):

lib/bluetooth/peer_manager/modules/security_dispatcher.c
```diff
@@ -245,16 +245,16 @@ static void sec_proc_start(uint16_t conn_handle, bool success, enum pm_conn_sec_
...
 static uint32_t link_secure_central_encryption(uint16_t conn_handle, uint16_t peer_id)
 {
-	struct pm_peer_data_const peer_data;
+	struct pm_peer_data peer_data;
 	uint32_t err_code;
 	const ble_gap_enc_key_t *existing_key = NULL;
 	bool lesc = false;
-	struct pm_peer_data_bonding bonding_data = { 0 };
+	struct pm_peer_data_bonding bonding_data = {};
+	const uint32_t buf_size = sizeof(struct pm_peer_data_bonding);
```
lemrey: Why is this a constant if it's passed by reference to `pds_peer_data_read()`?
Does that function not modify the size?

eivindj-nordic: This is the signature of `pds_peer_data_read`:
`uint32_t pds_peer_data_read(uint16_t peer_id, enum pm_peer_data_id data_id, struct pm_peer_data *const data, const uint32_t *const buf_len)`

so `buf_len` is constant.

anhmolt: In nRF5 SDK this could would allow setting `buf_size` to NULL to only have the pointer to memory being copied, but this is not supported with the new file system. So some cleanup is in order here I think.

eivindj-nordic (approved):

PR #466, opened 2025-11-10
softdevice_handler: cleanup irq_connect alternative application handler names

anhmolt:
- The RNG application handler function was not used and have been removed.
- Handler names are updated to better match the nrfx names and the names from `nrf_sd_isr.h`.

### DIFF
--- a/subsys/softdevice_handler/irq_connect.c
+++ b/subsys/softdevice_handler/irq_connect.c
@@ -79,37 +79,37 @@ __attribute__((weak)) void C_HardFault_Handler(void)
 	__asm__("SVC 255");
 }
 
-__attribute__((weak)) void C_TIMER0_Handler(void)
+__attribute__((weak)) void C_TIMER10_Handler(void)
 {
 	__asm__("SVC 255");
 }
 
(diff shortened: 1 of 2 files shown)


MirkoCovizzi (approved):

boards/nordic/bm_nrf54l15dk/bm_nrf54l15dk_nrf54l05_cpuapp_s115_softdevice_defconfig
```diff
@@ -87,7 +87,6 @@ CONFIG_NRFX_SPIM30=y
 CONFIG_NRFX_SYSTICK=y
 CONFIG_NRFX_TEMP=y
 CONFIG_NRFX_TIMER00=y
-CONFIG_NRFX_TIMER10=y
```
eivindj-nordic: I think we should keep this entry here. If disabled, we should set it to `n` with a comment.

eivindj-nordic (approved):

nordicjm (approved):

PR #536, opened 2025-12-02
Remove RRAM backend support in storage samples and enable SoftDevice as default in all samples

eivindj-nordic:


lemrey (commented):
This is not a good way to fix the issue. The issue is we do not have a board without SoftDevice. Therefore it does not make sense that any sample would try to disable `CONFIG_SOFTDEVICE` or support operation without `CONFIG_SOFTDEVICE`.

Here I think we should just remove the RRAM backend support from these samples.

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -292,6 +292,12 @@ DFU samples
 
 * Moved the MCUmgr samples to the :file:`applications/firmware_loader` folder.
 
+Subsystem samples
+-----------------
+
+* Moved the :ref:`bm_storage_sample` sample and the :ref:`bm_zms_sample` sample to the :file:`samples/subsys` sample folder.
```
```suggestion
* Moved the :ref:`bm_storage_sample` and :ref:`bm_zms_sample` samples to the :file:`samples/subsys` folder.
```

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -292,6 +292,12 @@ DFU samples
...
+* Removed the RRAM storage backend from the :ref:`bm_storage_sample` sample and the :ref:`bm_zms_sample` sample.
```
```suggestion
* Removed the RRAM storage backend from the :ref:`bm_storage_sample` and :ref:`bm_zms_sample` samples.
```

peknis (approved):
Approved with a couple of nits.

lemrey (approved):

nvlsianpu (approved):

PR #598, opened 2026-01-06
boards: nordic: bm_nrf54l15dk: Fix reserved memory location

nordicjm:
Moves these to the root node instead of being under the soc node


anhmolt (commented):
One comment, otherwise looks good 👍

anhmolt (approved):

kylebonnici (changes requested):
See https://github.com/nrfconnect/sdk-nrf-bm/pull/598#discussion_r2665292952

kylebonnici (approved):
I see that the node address was also dropped ignore my previous comment

eivindj-nordic (approved):
@nordicjm Please add changelog entry or remove changelog entry required label if you don't see a need for it.

PR #657, opened 2026-02-23
firmware_loader: merge SoC specific configs into prj.conf

anhmolt:
The SoC specific configurations are equal so they can be moved into the prj.conf file. If SoC specific configurations for the firmware loader is desired in the future, it can be reintroduced.


lemrey (approved):
Changelog entry not needed imo

eivindj-nordic (approved):

nvlsianpu (approved):

de-nordic (approved):

PR #661, opened 2026-02-24
boards: nrf54l15dk: disable Peripheral Resource Sharing (PRS) boxes

eivindj-nordic:
Disable all Peripheral Resource Sharing (PRS) boxes for nRF54l15DK.


lemrey (commented):
What's the reason for this change?
And also, why don't we disable all drivers by default?

anhmolt (approved):

lemrey:
> > What's the reason for this change?
> 
> Reason is to align with #621. The PRS boxes are for advanced use.

But a peripheral being for "advanced use" is not a clear and well defined characteristic. How do you define that? Is it really something we want to use to selectively enable/disable a peripheral at the board level?

> > And also, why don't we disable all drivers by default?
> 
> The reason to have the drivers enabled by default is to reduce the complexity for the user, and for the applications. We could remove it, though then we would need to enable it in all samples using the peripherals. Also, we might need to add board specific configuration files to some of the samples as not all SoCs have the same peripherals.

Perhaps we could think of peripherals that all samples want to use, like GPIO and UART and enable those, leaving the rest disabled.

nordicjm (approved):

b-gent (approved):

PR #683, opened 2026-03-06
lib: bluetooth: ble_gq: fix assert in req_queue_purge_schedule

eivindj-nordic:
The assert in req_queue_purge_schedule should trigger when the index is equal to the number of max connections, not when there is space in the list.


anhmolt (approved):
Thanks! I will triple check the logic of asserts I write in the future.

marenges (approved):

PR #687, opened 2026-03-11
lib: ble_db_discovery: improvements part two

anhmolt:
* fixed an issue where internal state indices were not properly reset to zero on subsequent discovery starts.
* Significantly reduce db_discovery instance memory usage by changing the discovery complete result to use pass by reference instead of pass by value.
* Improvements to db_discovery documentation.
* Update library maturity to support (from experimental).

test_bm_samples: PR-90

### DIFF
--- a/doc/nrf-bm/release_notes/release_notes_changelog.rst
+++ b/doc/nrf-bm/release_notes/release_notes_changelog.rst
@@ -151,6 +151,8 @@ Libraries
    * Updated:
 
       * The ``params`` union field of the :c:struct:`ble_db_discovery_evt` structure to an anonymous union.
+      * The :c:enumerator:`BLE_DB_DISCOVERY_COMPLETE` event to pass the service discovery result by-reference instead of by-value.
+        This removes one copy operation for each :c:enumerator:`BLE_DB_DISCOVERY_COMPLETE` event and significantly reduces the size of both :c:struct:`ble_db_discovery_evt` and :c:struct:`ble_db_discovery` structures.
 
    * Fixed:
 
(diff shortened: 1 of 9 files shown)


doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -108,7 +108,23 @@ Libraries
 
 * :ref:`lib_ble_db_discovery`:
```
```suggestion
* :ref:`lib_ble_db_discovery` library:
```

peknis (changes requested):

include/bm/bluetooth/ble_db_discovery.h
```diff
@@ -107,11 +107,11 @@ typedef void (*ble_db_discovery_evt_handler)(struct ble_db_discovery *db_discove
  */
 struct ble_db_discovery_config {
 	/**
-	 * @brief Event handler to be called by the DB Discovery library.
+	 * @brief Event handler to be called by the DB Discovery library (Required).
```
eivindj-nordic: Is this `(Required)` required? If we do this we should probably do that everywhere. I would have seen the need if there was a mix of required and optional in the struct.

eivindj-nordic (approved):
Some minor comments. Looks good to me when @peknis is happy with the documentation.

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -108,7 +108,23 @@ Libraries
...
+      * The library documentation page with more information.
+
+   * Fixed:
+
+      * An issue where starting a second round of discovery, after either a disconnect or a completed discovery, could result in incorrect internal state.
+        This could eventually lead to insufficient memory for discovering new services.
+
+   * Removed:
+
+      * The :c:struct:`ble_db_discovery_user_evt` structure after a rework.
```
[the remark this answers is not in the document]
anhmolt: Good catch! Thanks.

peknis (changes requested):

peknis (approved):

marenges (approved):

PR #690, opened 2026-03-12
lib: bluetooth: ble_adv: remove extra sd_ble_gap_adv_set_configure call

eivindj-nordic:
In most cases sd_ble_gap_adv_set_configure is called again after flags are set. Remove the call in flag_set and add to
ble_adv_restart_without_allow_list, which is the only place flag_set is called without calling sd_ble_gap_adv_set_configure after.

### DIFF
--- a/tests/unit/lib/bluetooth/ble_adv/src/unity_test.c
+++ b/tests/unit/lib/bluetooth/ble_adv/src/unity_test.c
@@ -370,9 +370,6 @@ static uint32_t stub_sd_ble_gap_adv_set_configure_directed_hd_success(
 
 	switch (cmock_num_calls) {
 	case 0:
-#if defined(CONFIG_BLE_ADV_USE_ALLOW_LIST)
-	case 1:
-#endif /* CONFIG_BLE_ADV_USE_ALLOW_LIST */
 #if defined(CONFIG_BLE_ADV_DIRECTED_ADVERTISING_HIGH_DUTY)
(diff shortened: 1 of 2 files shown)


tests/unit/lib/bluetooth/ble_adv/src/unity_test.c
```diff
@@ -1308,11 +1269,7 @@ void test_ble_adv_restart_without_allow_list_slow_success(void)
 
 	if (IS_ENABLED(CONFIG_BLE_ADV_SLOW_ADVERTISING)) {
 		if (IS_ENABLED(CONFIG_BLE_ADV_USE_ALLOW_LIST)) {
-			/* When using allow list, the adv flags in the AD data must be updated.
-			 * This is done with an additional call to sd_ble_gap_adv_set_configure().
-			 * Expect the function to be called twice if allow list is enabled.
-			 */
-			TEST_ASSERT_EQUAL(2, stub_sd_ble_gap_adv_set_configure_num_calls);
+			TEST_ASSERT_EQUAL(1, stub_sd_ble_gap_adv_set_configure_num_calls);
```
anhmolt: This assert check is now the same as that in the else block and can be moved outside of the if-else.

This applies to more, if not most, of the `stub_sd_ble_gap_adv_set_configure_num_calls` assert checks that have been changed in this commit.

anhmolt (commented):
I like how this seems to clean up the unit tests a bit. See comments.

anhmolt (changes requested):

anhmolt (approved):

marenges (approved):

PR #721, opened 2026-03-27
samples: bluetooth: ble_hrs_peripheral_central: Add sample

sondrep:
Adds the hrs_peripheral_central sample.


subsys/bluetooth/services/ble_hrs/hrs.c
```diff
@@ -192,23 +192,25 @@ void ble_hrs_on_ble_evt(const ble_evt_t *ble_evt, void *hrs_instance)
 {
 	__ASSERT(ble_evt, "BLE event is NULL");
 	__ASSERT(hrs_instance, "BLE instance is NULL");
+	const ble_gap_evt_t *gap_evt = &ble_evt->evt.gap_evt;
 
+	if (gap_evt->params.connected.role != BLE_GAP_ROLE_PERIPH){
+		return;
+	}
```
eivindj-nordic: Commit message should describe what it is fixing.

subsys/bluetooth/services/ble_hrs/hrs.c
```diff
@@ -192,23 +192,25 @@ void ble_hrs_on_ble_evt(const ble_evt_t *ble_evt, void *hrs_instance)
...
 		break;
-
 	case BLE_GATTS_EVT_WRITE:
 		on_write(hrs_instance, &ble_evt->evt.gatts_evt);
 		break;
-
 	default:
 		break;
 	}
+
```
eivindj-nordic: nit: remove

subsys/bluetooth/services/ble_hrs_client/ble_hrs_client.c
```diff
@@ -152,7 +154,16 @@ void ble_hrs_on_db_disc_evt(struct ble_hrs_client *hrs_client,
...
-			break;
+		}
+		else if (db_char->characteristic.uuid.uuid == BLE_UUID_BODY_SENSOR_LOCATION_CHAR) {
+			hrs_client_evt.discovery_complete.handles.bsl_cccd_handle = db_char->cccd_handle;
+			hrs_client_evt.discovery_complete.handles.bsl_handle = db_char->characteristic.handle_value;
+			ble_gatts_value_t val = {
+				.p_value = &hrs_client_evt.discovery_complete.body_sensor_location,
+				.len = sizeof(hrs_client_evt.discovery_complete.body_sensor_location),
+			};
+			//LOG_INF("sd_ble_gatts_value_get returned %#x, location = %u" ,sd_ble_gatts_value_get(db_discovery_evt->conn_handle, db_char->characteristic.handle_value, &val), hrs_client_evt.discovery_complete.body_sensor_location);
```
eivindj-nordic: nit: remove

subsys/bluetooth/services/ble_hrs_client/ble_hrs_client.c
```diff
@@ -194,6 +205,27 @@ uint32_t ble_hrs_client_init(struct ble_hrs_client *hrs_client,
 	return ble_db_discovery_service_register(hrs_client_config->db_discovery, &hrs_uuid);
 }
 
+static void on_read_rsp(struct ble_hrs_client *hrs_client, const ble_evt_t *ble_evt) {
+
+	const ble_gattc_evt_read_rsp_t *read_rsp = &ble_evt->evt.gattc_evt.params.read_rsp;
+	struct ble_hrs_client_evt evt = {.conn_handle = ble_evt->evt.gattc_evt.conn_handle, .evt_type = BLE_HRS_CLIENT_EVT_BSL_UPDATE, .body_sensor_location = ble_evt->evt.gattc_evt.params.read_rsp.data[0]};
```
eivindj-nordic: fix formatting

subsys/bluetooth/services/ble_hrs_client/ble_hrs_client.c
```diff
@@ -194,6 +205,27 @@ uint32_t ble_hrs_client_init(struct ble_hrs_client *hrs_client,
...
+
+	if (hrs_client->conn_handle != ble_evt->evt.gattc_evt.conn_handle) {
+		return;
+	}
+
+	/* Check if this is a heart rate notification. */
```
eivindj-nordic: Update comment

subsys/bluetooth/services/ble_hrs_client/ble_hrs_client.c
```diff
@@ -204,8 +236,15 @@ void ble_hrs_client_on_ble_evt(const ble_evt_t *ble_evt, void *hrs_client)
 		on_hvx(hrs_client, ble_evt);
 		break;
 	case BLE_GAP_EVT_DISCONNECTED:
+		const ble_gap_evt_t *gap_evt = &ble_evt->evt.gap_evt;
+		if (gap_evt->params.connected.role != BLE_GAP_ROLE_CENTRAL){
+			return;
+		}
```
eivindj-nordic: Move to start of event handler?

samples/bluetooth/ble_hrs_peripheral_central/sample.yaml
```diff
@@ -0,0 +1,37 @@
...
+    platform_allow:
+      - bm_nrf54l15dk/nrf54l05/cpuapp/s145_softdevice
+      - bm_nrf54l15dk/nrf54l05/cpuapp/s145_softdevice/mcuboot
+      - bm_nrf54l15dk/nrf54l10/cpuapp/s145_softdevice
+      - bm_nrf54l15dk/nrf54l10/cpuapp/s145_softdevice/mcuboot
+      - bm_nrf54l15dk/nrf54l15/cpuapp/s145_softdevice
+      - bm_nrf54l15dk/nrf54l15/cpuapp/s145_softdevice/mcuboot
+      - bm_nrf54lm20dk/nrf54lm20a/cpuapp/s145_softdevice
+      - bm_nrf54ls05dk/nrf54ls05b/cpuapp/s145_softdevice
+      - bm_nrf54lv10dk/nrf54lv10a/cpuapp/s145_softdevice
```
eivindj-nordic: MCUboot variant for these can be added now.

samples/bluetooth/ble_hrs_peripheral_central/sample.yaml
```diff
@@ -0,0 +1,37 @@
...
+    extra_configs:
+      - CONFIG_BM_UARTE_CONSOLE_UARTE_USE_HWFC=y
+      - CONFIG_BM_UARTE_CONSOLE_UARTE_PARITY_INCLUDED=y
+    tags: ci_build
```
eivindj-nordic: I think this configuration can be removed here

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,125 @@
...
+
+The Heart Rate Service Peripheral Central sample demonstrates how you can implement the Heart Rate profile as a server and client using |BMlong|.
+
+Requirements
+************
+
+The sample supports the following development kits:
+
+.. tabs::
+
```
eivindj-nordic: We have a template that can be used here, see other central samples.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,125 @@
...
+Button 1:
+  Press to disconnect from the connected peripheral device.
+
+  Keep the button pressed while resetting the board to delete bonding information for all peers stored on the device.
+
+Button 2:
+  Press to disconnect from the connected client device.
+
+  Keep the button pressed while resetting the board to delete bonding information for all peers stored on the device.
```
eivindj-nordic: This is covered by button 1(?)

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,125 @@
...
+Button 0:
+  Press to disable allow list.
+
+Button 1:
+  Press to disconnect from the connected peripheral device.
+
+  Keep the button pressed while resetting the board to delete bonding information for all peers stored on the device.
+
+Button 2:
+  Press to disconnect from the connected client device.
```
eivindj-nordic: client and server, or peripheral and central.
```suggestion
  Press to disconnect from the connected central device.
```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,125 @@
...
+********************
+
+This sample can be found under :file:`samples/bluetooth/ble_hrs_peripheral_central/` in the |BMshort| folder structure.
+
+For details on how to create, configure, and program a sample, see :ref:`getting_started_with_the_samples`.
+
+Testing
+=======
+
+This sample requires three devices to test, one running this sample, another one running the :ref:`ble_hrs_sample` sample, and another running the :ref:`ble_hrs_central_sample`.
```
eivindj-nordic: Or, as many might want, a phone acting as the second central

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Complete the following steps to test the sample:
+
+1. Compile and program the application.
+#. In the Serial Terminal, observe that the ``Advertising as nRF_BM_HRS_Hub`` message is printed.
```
eivindj-nordic: Not sure if Bridge is the sample name we want to use for it, though this should match the name of the sample (if using bridge, update peripheral_central -> bridge in folder name, documentation etc.) 
```suggestion
#. In the Serial Terminal, observe that the ``Advertising as nRF_BM_HRS_Bridge`` message is printed.
```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,125 @@
...
+   You can configure the advertising name using the :kconfig:option:`CONFIG_SAMPLE_BLE_DEVICE_NAME` Kconfig option.
+   For information on how to do this, see `Configuring Kconfig`_.
+#. Observe that the ``BLE HRS Peripheral Central sample initialized`` message is printed.
```
eivindj-nordic: We should put this log before the advertising log, swap order in the source file as well.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,125 @@
...
+#. Program the second development kit with the :ref:`ble_hrs_sample` sample.
+#. Program the third development kit with the :ref:`ble_hrs_central_sample` sample, set the :kconfig:option:`CONFIG_SAMPLE_TARGET_PERIPHERAL_NAME` Kconfig option to ``nRF_BM_HRS_Peripheral_Central``, and set the :kconfig:option:`CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_NAME` Kconfig option to ``y``.
+#. Observe that the ``Scan filter match`` message is printed, followed by ``Connecting to target`` and ``Connected``, this will happen twice, once for each peer.
```
eivindj-nordic: We should have only one scan filter match(?). The other central device is not advertising.

samples/bluetooth/ble_hrs_peripheral_central/Kconfig
```diff
@@ -0,0 +1,43 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
```
eivindj-nordic: 2026 Here and others

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+#include <bm/bluetooth/ble_gq.h>
+#include <bm/bluetooth/services/ble_bas.h>
+#include <bm/bluetooth/services/ble_bas_client.h>
+#include <bm/bluetooth/services/ble_hrs.h>
+#include <bm/bluetooth/services/ble_hrs_client.h>
+#include <bm/bluetooth/ble_common.h>
+#include <bm/bluetooth/services/uuid.h>
+#include <bm/softdevice_handler/nrf_sdh.h>
+#include <bm/softdevice_handler/nrf_sdh_ble.h>
+#include <bm/softdevice_handler/nrf_sdh_soc.h>
```
eivindj-nordic: This is not sorted. Are all required?

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+
+#include <zephyr/bluetooth/uuid.h>
```
eivindj-nordic: We should have our internal UUID header(?)

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+/* Gatt queue instance. */
+BLE_GQ_DEF(ble_gq);
+/* BLE heart rate service instance */
+BLE_HRS_DEF(ble_hrs);
+/* Structure used to identify the heart rate client module. */
+BLE_HRS_CLIENT_DEF(ble_hrs_client);
+/* Scanning module instance. */
+BLE_SCAN_DEF(ble_scan);
+
+/* Current connection handle. */
```
eivindj-nordic:
```suggestion
/* Current connection handles. */
```

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+			if (nrf_err != 0) {
+				LOG_ERR("db discovery start failed, nrf_error %#x", nrf_err);
+			}
+			nrf_gpio_pin_write(BOARD_PIN_LED_1, BOARD_LED_ACTIVE_STATE);
+		} else if (gap_evt->params.connected.role == BLE_GAP_ROLE_PERIPH){
+			central_conn_handle = ble_evt->evt.gap_evt.conn_handle;
+			nrf_gpio_pin_write(BOARD_PIN_LED_2, BOARD_LED_ACTIVE_STATE);
+		}
+
+		if (gap_evt->params.connected.role == BLE_GAP_ROLE_CENTRAL) {
```
eivindj-nordic: group with if above

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+		break;
+
+	case BLE_GAP_EVT_DISCONNECTED:
+		LOG_INF("Disconnected, reason %#x", gap_evt->params.disconnected.reason);
+
+		if (central_conn_handle == gap_evt->conn_handle) {
+			central_conn_handle = BLE_CONN_HANDLE_INVALID;
+			nrf_gpio_pin_write(BOARD_PIN_LED_2, !BOARD_LED_ACTIVE_STATE);
+		}
+		if (peripheral_conn_handle == gap_evt->conn_handle) {
```
eivindj-nordic:
```suggestion
		} else if (peripheral_conn_handle == gap_evt->conn_handle) {
```

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+			scan_start(false);
+		}
+
+		break;
+
+	case BLE_GAP_EVT_DISCONNECTED:
+		LOG_INF("Disconnected, reason %#x", gap_evt->params.disconnected.reason);
+
+		if (central_conn_handle == gap_evt->conn_handle) {
+			central_conn_handle = BLE_CONN_HANDLE_INVALID;
```
eivindj-nordic: Clear central_conn here, if not `active_conn_count` is still 1 after disconnect?

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+			nrf_gpio_pin_write(BOARD_PIN_LED_2, !BOARD_LED_ACTIVE_STATE);
+		}
+		if (peripheral_conn_handle == gap_evt->conn_handle) {
+			peripheral_conn_handle = BLE_CONN_HANDLE_INVALID;
+			atomic_clear_bit(&central_conn, idx);
+			nrf_gpio_pin_write(BOARD_PIN_LED_1, !BOARD_LED_ACTIVE_STATE);
+		}
+
+		if (active_conn_count(&central_conn) < CONFIG_NRF_SDH_BLE_CENTRAL_LINK_COUNT) {
+			scan_start(false);
```
eivindj-nordic: If we get a disconnect event with central_conn_handle, we know a central is disconnected and we can start scanning.

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+		.lesc = 1,
+		.keypress = 0,
+		.io_caps = BLE_GAP_IO_CAPS_NONE,
+		.oob = 0,
+		.min_key_size = 7,
+		.max_key_size = 16,
+		.kdist_own.enc = 1,
+		.kdist_own.id = 1,
+		.kdist_peer.enc = 1,
+		.kdist_peer.id = 1,
```
eivindj-nordic:
```suggestion
		.kdist_own = {
			.enc = 1,
			.id = 1,
		},
		.kdist_peer = {
			.enc = 1,
			.id = 1,
		},
```

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+		{.uuid = BLE_UUID_HEART_RATE_SERVICE, .type = BLE_UUID_TYPE_BLE},
+	};
+	struct ble_adv_config ble_adv_cfg = {
+		.conn_cfg_tag = CONFIG_NRF_SDH_BLE_CONN_TAG,
+		.evt_handler = ble_adv_evt_handler,
+		.adv_data = {.name_type = BLE_ADV_DATA_FULL_NAME,
+			     .flags = BLE_GAP_ADV_FLAGS_LE_ONLY_GENERAL_DISC_MODE},
+		.sr_data.uuid_lists.complete = {.len = ARRAY_SIZE(adv_uuid_list),
+						.uuid = &adv_uuid_list[0]},
+	};
```
eivindj-nordic: Fix formatting, here and others

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+
+	nrf_err = ble_adv_init(&ble_adv, &ble_adv_cfg);
+
+	return nrf_err;
```
eivindj-nordic:
```suggestion
	return ble_adv_init(&ble_adv, &ble_adv_cfg);
```

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+	}
+
+	LOG_INF("Button disconnect peripheral");
+
+	uint32_t nrf_err = sd_ble_gap_disconnect(peripheral_conn_handle,
+						 BLE_HCI_REMOTE_USER_TERMINATED_CONNECTION);
+
+	if (nrf_err != 0) {
+		LOG_ERR("ble gap disconnect failed, nrf_error %#x", nrf_err);
+	}
```
eivindj-nordic:
```suggestion
	uint32_t nrf_err ;
	
	if (peripheral_conn_handle == BLE_CONN_HANDLE_INVALID || action != BM_BUTTONS_PRESS) {
		return;
	}

	LOG_INF("Button disconnect peripheral");

	nrf_err = sd_ble_gap_disconnect(peripheral_conn_handle, BLE_HCI_REMOTE_USER_TERMINATED_CONNECTION);
	if (nrf_err) {
		LOG_ERR("ble gap disconnect failed, nrf_error %#x", nrf_err);
	}
```

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+{
+
+	uint32_t nrf_err;
+
+	switch (evt->evt_type) {
+	case BLE_HRS_CLIENT_EVT_DISCOVERY_COMPLETE:
+		ble_hrs_body_sensor_location_set(&ble_hrs, evt->discovery_complete.body_sensor_location);
+
+		nrf_err = ble_hrs_client_handles_assign(hrs, evt->conn_handle, &evt->discovery_complete.handles);
+		if (nrf_err != 0) {
```
eivindj-nordic: For errno we have 0 for success, for nrf_errors we have the define NRF_SUCCESS, so comparing nrf_err with 0 is not right. Though, as NRF_SUCCESS is defined as 0 in our case we allow to do
```suggestion
		if (nrf_err) {
```
alternatively it should be 
```suggestion
		if (nrf_err != NRF_SUCCESS) {
```

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+			}
+			rr_avg = rr_avg / evt->hrm_notification.rr_intervals_cnt;
+			LOG_INF("rr_interval (avg) = %d.", rr_avg);
+		}
+		nrf_err = ble_hrs_heart_rate_measurement_send(&ble_hrs, evt->hrm_notification.hr_value);
+		if (nrf_err) {
+		/* Ignore if not connected, or CCCD not written/configured by peer. */
+			if (nrf_err != BLE_ERROR_INVALID_CONN_HANDLE &&
+		    	nrf_err != NRF_ERROR_INVALID_STATE &&
+		    	nrf_err != BLE_ERROR_GATTS_SYS_ATTR_MISSING) {
```
eivindj-nordic: Alignment is off, here and others

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+		if (nrf_err != 0) {
+			LOG_ERR("ble_bas_client_handles_assign failed, nrf_error %#x", nrf_err);
+		}
+
+		/* Battery service discovered. Enable notification of battery level. */
+		nrf_err = ble_bas_client_bl_notif_enable(bas);
+		if (nrf_err != 0) {
+			LOG_ERR("ble_bas_client_bl_notif_enable failed, nrf_error %#x", nrf_err);
+		}
+	break;
```
eivindj-nordic:
```suggestion
		break;
```

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+	uint32_t nrf_err = ble_conn_params_evt_handler_set(conn_params_evt_handler);
+
+	if (nrf_err) {
+		LOG_ERR("ble_conn_params_evt_handler_set failed, nrf_error %#x", nrf_err);
+	}
+
+	return nrf_err;
+}
+
+static void on_conn_params_evt(const struct ble_conn_params_evt *evt)
```
eivindj-nordic: Lets group all event handlers in the file, and have all init functions afterwards.

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,965 @@
...
+	scan_start(erase_bonds);
+
+	nrf_gpio_pin_write(BOARD_PIN_LED_0, BOARD_LED_ACTIVE_STATE);
+	LOG_INF("BLE HRS Central sample initialized");
+
+	while (true) {
+#if defined(CONFIG_PM_LESC)
+		(void)nrf_ble_lesc_request_handler();
+#endif
+idle:
```
eivindj-nordic: this can be moved before the while loop.

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
sondrep: The naming convention might be a bit confusing here, central_conn is the number of connections where the device is acting as a central while the central_conn_handle is the connection handle for the peer that is connected as a central.

subsys/bluetooth/services/ble_hrs_client/ble_hrs_client.c
asilz: I put it there because the check is dependent on the fact that the event is a gap event. I felt that it would be unsafe to access gap event members if it was a gattc event.

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,953 @@
+/*
+ * Copyright (c) 2014-2026 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <bm/bluetooth/ble_adv.h>
+#include <bm/bluetooth/ble_scan.h>
```
PizzaAllTheWay:
```suggestion
#include <bm/bluetooth/ble_scan.h>
#include <bm/bluetooth/peer_manager/nrf_ble_lesc.h>
```

Library implicit usage fix

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,953 @@
...
+
+static void pm_evt_handler(const struct pm_evt *evt)
+{
+	pm_handler_on_pm_evt(evt);
+	pm_handler_disconnect_on_sec_failure(evt);
+	pm_handler_flash_clean(evt);
+
+	switch (evt->evt_id) {
+	case PM_EVT_PEERS_DELETE_SUCCEEDED:
+		scan_start(false);
```
PizzaAllTheWay:
```suggestion
		scan_start(false);
		advertising_start(false);
```
Bug: if `erase_bonds` activated at startup, device is left in non-advertising state, fix

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,953 @@
...
+	filter_data.name_filter.name = CONFIG_SAMPLE_TARGET_PERIPHERAL_NAME;
+
+	nrf_err = ble_scan_filter_add(&ble_scan, BLE_SCAN_NAME_FILTER, &filter_data);
+	if (nrf_err) {
+		LOG_ERR("nrf_ble_scan_filter_add name failed, nrf_error %#x", nrf_err);
+	}
+#endif /* CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_NAME */
+
+#if defined(CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_ADDR)
+	filter_data.addr_filter.addr = target_periph_addr;
```
PizzaAllTheWay:
```suggestion
	filter_data.addr_filter.addr = (uint8_t *)target_periph_addr;
```
Fix compile warning: cast away `const` from `target_periph_addr` to match the non-const `addr_filter.addr` field type.

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,953 @@
...
+		},
+		{
+			.pin_number = BOARD_PIN_BTN_2,
+			.active_state = BM_BUTTONS_ACTIVE_LOW,
+			.pull_config = BM_BUTTONS_PIN_PULLUP,
+			.handler = button_handler_disconnect_central,
+		},
+	};
+
+	LOG_INF("BLE HRS Peripheral Central sample initialized.");
```
PizzaAllTheWay:
```suggestion
	LOG_INF("BLE HRS Peripheral Central sample started.");
```

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,953 @@
...
+	nrf_err = scan_init();
+	if (nrf_err) {
+		LOG_ERR("Failed to initialize scan library, nrf_error %#x", nrf_err);
+		goto idle;
+	}
+
+	scan_start(erase_bonds);
+
+	nrf_gpio_pin_write(BOARD_PIN_LED_0, BOARD_LED_ACTIVE_STATE);
+	LOG_INF("BLE HRS Central sample initialized");
```
PizzaAllTheWay:
```suggestion
	LOG_INF("BLE HRS Peripheral Central sample initialized");
```

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,953 @@
...
+	filter_data.addr_filter.addr = target_periph_addr;
+	nrf_err = ble_scan_filter_add(&ble_scan, BLE_SCAN_ADDR_FILTER, &filter_data);
+	if (nrf_err) {
+		LOG_ERR("nrf_ble_scan_filter_add address failed, nrf_error %#x", nrf_err);
+	}
+#endif /* CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_ADDR */
+
+	nrf_err = ble_scan_filters_enable(
+		&ble_scan, BLE_SCAN_UUID_FILTER | BLE_SCAN_NAME_FILTER | BLE_SCAN_ADDR_FILTER,
+		false);
```
PizzaAllTheWay:
```suggestion
	uint8_t filter_mode_mask = BLE_SCAN_UUID_FILTER;
#if defined(CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_NAME)
    	filter_mode_mask |= BLE_SCAN_NAME_FILTER;
#endif
#if defined(CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_ADDR)
    	filter_mode_mask |= BLE_SCAN_ADDR_FILTER;
#endif

	nrf_err = ble_scan_filters_enable(&ble_scan, filter_mode_mask, true);
```
The current call uses `match_all = false` and unconditionally enables the NAME and ADDR filters. This has two problems: with `match_all = false` the UUID match alone is enough, so the address and name filters are silently bypassed and the Kconfig options have no effect. If the flag is flipped to `true`, the NAME and ADDR filters are still enabled even when their Kconfig options are disabled, so the empty filters block every connection. The fix is to build the enable mask dynamically from the active Kconfig options and use `match_all = true` so that strict matching only applies to filters the user actually configured. This also makes the filter feature more intuitive and easier to showcase, since enabling an address or name filter in Kconfig will now produce the expected strict matching behaviour.

samples/bluetooth/ble_hrs_peripheral_central/prj.conf
```diff
@@ -0,0 +1,55 @@
...
+
+# Advertising library
+CONFIG_BLE_ADV=y
+
+# Peer manager
+CONFIG_PEER_MANAGER=y
+CONFIG_PM_LESC=y
+CONFIG_BM_ZMS=y
+
+CONFIG_BLE_GQ_MAX_CONNECTIONS=2
```
PizzaAllTheWay:
```suggestion
# GATT queue: one queue per concurrent connection
CONFIG_BLE_GQ_MAX_CONNECTIONS=2
```
Add comment for clarity, something like this.

samples/bluetooth/ble_hrs_peripheral_central/prj.conf
```diff
@@ -0,0 +1,55 @@
...
+CONFIG_SOFTDEVICE=y
+
+# SoftDevice handler link counts
+CONFIG_NRF_SDH_BLE_TOTAL_LINK_COUNT=2
+CONFIG_NRF_SDH_BLE_CENTRAL_LINK_COUNT=1
+
+# Enable RNG
+CONFIG_NRF_SECURITY=y
+CONFIG_MBEDTLS_PSA_CRYPTO_C=y
+CONFIG_PSA_WANT_GENERATE_RANDOM=y
```
PizzaAllTheWay:
```suggestion
# Enable RNG
CONFIG_NRF_SECURITY=y
CONFIG_MBEDTLS_PSA_CRYPTO_C=y
CONFIG_PSA_WANT_GENERATE_RANDOM=y

# Enable Crypto functionality required by LE Secure Connections pairing (ECDH over NIST P-256)
CONFIG_PSA_WANT_ALG_ECDH=y
CONFIG_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE=y
CONFIG_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT=y
CONFIG_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT=y
CONFIG_PSA_WANT_ECC_SECP_R1_256=y

# PSA key storage: one slot per concurrent pairing
CONFIG_MBEDTLS_PSA_STATIC_KEY_SLOTS=y
CONFIG_MBEDTLS_PSA_KEY_SLOT_COUNT=2
```

Without these crypto options, `CONFIG_PM_LESC=y` is silently disabled and LESC pairing fails at runtime.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,88 @@
...
+  Keep the button pressed while resetting the board to delete bonding information for all peers stored on the device.
+
+Button 2:
+  Press to disconnect from the connected central device.
+
+LED 0:
+   Lit when the device is initialized.
+
+LED 1:
+   Lit when a peripheral device is connected.
```
PizzaAllTheWay:
```suggestion
   Lit when connected to a peripheral device.
```
Maybe something like this instead?

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,88 @@
...
+LED 2:
+   Lit when a central device is connected.
+
+Building and running
+********************
+
+This sample can be found under :file:`samples/bluetooth/ble_hrs_peripheral_central/` in the |BMshort| folder structure.
+
+For details on how to create, configure, and program a sample, see :ref:`getting_started_with_the_samples`.
+
```
PizzaAllTheWay:
```suggestion
Scan filtering options
**********************

The sample always scans for devices advertising the Heart Rate Service UUID
(``0x180D``). Two optional filters can narrow this down further:

.. list-table::
   :header-rows: 1

   * - Kconfig option
     - Matches on
   * - :kconfig:option:`CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_NAME`
     - Advertised device name (eks: ``"MyDeviceName"``)
   * - :kconfig:option:`CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_ADDR`
     - Exact 48-bit Bluetooth address

Useful for:

* **Development**: Multiple HRS devices nearby, only want to connect to yours.
* **Production**: Bridge paired with one specific sensor, ignore everything else.
* **Bonded scenarios**: Combine with allow lists for tighter control.

Note: "peripheral" in the option name refers to the *remote* device being scanned
for, not this device's role.
```
Adds a new section documenting the optional scan filter Kconfigs, so users know these options exist and what they are for before running the sample.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,88 @@
...
+#. Observe that the ``BLE HRS Peripheral Central sample initialized`` message is printed.
+#. In the Serial Terminal, observe that the ``Advertising as nRF_BM_HRS_bridge`` message is printed.
+   You can configure the advertising name using the :kconfig:option:`CONFIG_SAMPLE_BLE_DEVICE_NAME` Kconfig option.
+   For information on how to do this, see `Configuring Kconfig`_.
+#. Program the second development kit with the :ref:`ble_hrs_sample` sample.
+#. Observe that the ``Scan filter match`` message is printed, followed by ``Connecting to target`` and ``Connected``, when connecting to the peripheral device.
+#. Observe that the ``Heart rate service discovered.`` message is printed.
+#. Connect to the device from nRF Connect (the device is advertising as "nRF_BM_HRS_bridge").
+#. Observe that the ``Connecting to target`` and ``Connected`` messages are printed when connecting to the central device.
+#. Observe that the device starts receiving heart rate measurement notifications and forwarding them to the central.
```
PizzaAllTheWay:
```suggestion
#. Observe that the device starts receiving heart rate measurement notifications and forwarding them to the central.
#. Note the address printed in the log when connecting, eks:

   .. code-block:: console

      Connecting to target AABBCCDDEEFF

#. Enable the address filter in Kconfig with that exact value:

   .. code-block:: cfg

      CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_ADDR=y
      CONFIG_SAMPLE_TARGET_PERIPHERAL_ADDR=0xAABBCCDDEEFF

   Rebuild and flash. Confirm the sample still connects to the same peripheral.

#. Change the address to a wrong value, rebuild and flash. Confirm the sample no
   longer connects to any peripheral.

#. Repeat the same flow with :kconfig:option:`CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_NAME`
   (easier, uses the advertised name).
```
Extend the testing steps to cover the optional scan filters, so users can verify the address and name filtering actually work as expected.

subsys/bluetooth/services/ble_hrs/hrs.c
```diff
@@ -193,6 +193,13 @@ void ble_hrs_on_ble_evt(const ble_evt_t *ble_evt, void *hrs_instance)
 	__ASSERT(ble_evt, "BLE event is NULL");
 	__ASSERT(hrs_instance, "BLE instance is NULL");
 
+	if (ble_evt->header.evt_id >= BLE_GAP_EVT_BASE &&
+	    ble_evt->header.evt_id <= BLE_GAP_EVT_LAST &&
+	    ble_evt->evt.gap_evt.params.connected.role != BLE_GAP_ROLE_PERIPH &&
+	    ble_evt->evt.gap_evt.params.connected.role != BLE_GAP_ROLE_INVALID) {
+		return;
+	}
```
PizzaAllTheWay:
```suggestion
	/* Filter:
	 * Only handle peripheral-role GAP events.
	 * Central-role events would otherwise overwrite hrs->conn_handle and break notifications
	 * to the real peer.
	 */
	if (ble_evt->header.evt_id >= BLE_GAP_EVT_BASE &&
	    ble_evt->header.evt_id <= BLE_GAP_EVT_LAST &&
	    ble_evt->evt.gap_evt.params.connected.role != BLE_GAP_ROLE_PERIPH &&
	    ble_evt->evt.gap_evt.params.connected.role != BLE_GAP_ROLE_INVALID) {
		return;
	}
```
Perhaps add a comment here for code clarity, because the role filter is not obvious at a glance. Wording is just a suggestion, feel free to adjust.

subsys/bluetooth/services/ble_hrs_client/ble_hrs_client.c
PizzaAllTheWay:
```suggestion
	/* Iterate discovered characteristics:
	 * - HRM:  store value handle and CCCD handle for later notification setup.
	 * - BSL:  store value handle and issue a Read Request; the value will
	 *         arrive asynchronously via BLE_GATTC_EVT_READ_RSP.
	 */
```
Update this comment for code clarity. The current wording only describes part of what the loop does, it also handles BSL and triggers an async read now. Wording is just a suggestion, feel free to adjust.

PizzaAllTheWay (changes requested):
Good job, core functionality works properly, very nice 😄 
A few changes that would be nice to see. Mostly small fixes, clarifying comments, and a couple of behaviour changes around bond erase and scan filters.

PizzaAllTheWay:
Also, add a line in the changelog noting that this sample was added, and that the BLE HRS CLIENT API changed with the addition of `BLE_HRS_CLIENT_EVT_BSL_UPDATE` in `ble_hrs_client_evt_type`

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
asilz: I feel like casting away constness could be dangerous and could lead to errors in the future. Do you think I should make `target_periph_addr` not const?

PizzaAllTheWay: Sounds like a good plan, const the filter fields at the API level is the right fix. I see you have already opened a PR for it (#791). Once that lands, this cast won't be needed at all and `filter_data.addr_filter.addr = target_periph_addr` will compile cleanly, no changes needed here then.

PizzaAllTheWay (approved):

subsys/bluetooth/services/ble_hrs/hrs.c
```diff
@@ -193,6 +193,18 @@ void ble_hrs_on_ble_evt(const ble_evt_t *ble_evt, void *hrs_instance)
...
 
+	/* Filter:
+	 * Only handle peripheral-role GAP events.
+	 * Central-role events would otherwise overwrite hrs->conn_handle and break notifications
+	 * to the real peer.
+	 */
+	if (ble_evt->header.evt_id >= BLE_GAP_EVT_BASE &&
+	    ble_evt->header.evt_id <= BLE_GAP_EVT_LAST &&
+	    ble_evt->evt.gap_evt.params.connected.role != BLE_GAP_ROLE_PERIPH &&
+	    ble_evt->evt.gap_evt.params.connected.role != BLE_GAP_ROLE_INVALID) {
```
anhmolt: `params.connected` should not be accessed here without first ensuring that `evt_id == BLE_GAP_EVT_CONNECTED`.

Also, I think there is no benefit in checking for `BLE_GAP_ROLE_INVALID`, SoftDevice will not return it for a `BLE_GAP_EVT_CONNECTED` event.

anhmolt (dismissed):

subsys/bluetooth/services/ble_hrs_client/ble_hrs_client.c
eivindj-nordic: Right, it is accessing the connected event parameters, so it should be kept here.

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
anhmolt: Removed `active_conn_count`. It is not required IMO when we only allow one central and one peripheral connection.

subsys/bluetooth/services/ble_hrs/hrs.c
anhmolt: Fixed

include/bm/bluetooth/services/ble_hrs_client.h
```diff
@@ -98,6 +104,8 @@ struct ble_hrs_client_evt {
 		} discovery_complete;
 		/** @ref BLE_HRS_CLIENT_EVT_HRM_NOTIFICATION event data. */
 		struct ble_hrs_measurement hrm_notification;
+		/** @ref BLE_HRS_CLIENT_EVT_BSL_UPDATE event data. */
+		uint8_t body_sensor_location;
```
eivindj-nordic: (discussed offline) struct to align with others?

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
```diff
@@ -0,0 +1,908 @@
...
+	switch (adv_evt->evt_type) {
+	case BLE_ADV_EVT_ERROR:
+		LOG_ERR("Advertising error %#x", adv_evt->error.reason);
+		break;
+	default:
+		break;
+	}
+}
+
+static void hrs_c_evt_handler(struct ble_hrs_client *hrs, const struct ble_hrs_client_evt *evt)
```
eivindj-nordic: nit: to align with others (or oposite)
```suggestion
static void ble_hrs_c_evt_handler(struct ble_hrs_client *hrs, const struct ble_hrs_client_evt *evt)
```

include/bm/bluetooth/services/ble_hrs_client.h
anhmolt: Updated.

samples/bluetooth/ble_hrs_peripheral_central/src/main.c
anhmolt: Aligned the handler names the other way, to `<srvc>_evt_handler` and `<srvc>_client_evt_handler`, skipping the `ble_` prefix.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,133 @@
...
+
+      .. include:: /includes/supported_boards_all_mcuboot_variants_s145.txt
+
+Overview
+********
+
+This sample advertises and scans for devices that advertise with the :ref:`lib_ble_service_hrs` UUID (0x180D) and initiates a connection when a device is found.
+When a peripheral device is connected, the sample starts the service discovery procedure.
+If this succeeds, the sample subscribes to the Heart Rate Measurement characteristic to receive heart rate notifications.
+If a central connects to this device the central subscribes to the Heart Rate Measurement characteristic to receive heart rate notifications forwarded from the peripheral device.
```
eivindj-nordic: I think this could use a rewrite. Perhaps something like. 


```suggestion
This sample acts simultaneously as both a peripheral and central device. 
* As a peripheral the sample advertises with the :ref:`lib_ble_service_hrs` UUID (0x180D). If a central connects to this device the central subscribes to the Heart Rate Measurement characteristic to receive heart rate notifications forwarded from the peripheral device.
* As a central the sample scans for other devices that advertise with the :ref:`lib_ble_service_hrs` UUID (0x180D). When a device is found it initiates a connection before starting the service discovery procedure and subscribing to the Heart Rate Measurement characteristic notifications.
```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,133 @@
...
+     - Advertised device name (eks: ``"MyDeviceName"``)
+   * - :kconfig:option:`CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_ADDR`
+     - Exact 48-bit Bluetooth address
+
+Useful for:
+* **Development**: Multiple HRS devices nearby, only want to connect to yours.
+* **Production**: Bridge paired with one specific sensor, ignore everything else.
+* **Bonded scenarios**: Combine with allow lists for tighter control.
+Note: "peripheral" in the option name refers to the *remote* device being scanned
+for, not this device's role.
```
eivindj-nordic: Do we need this?

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,133 @@
...
+   .. code-block:: cfg
+
+      CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_ADDR=y
+      CONFIG_SAMPLE_TARGET_PERIPHERAL_ADDR=0xAABBCCDDEEFF
+
+   Rebuild and flash. Confirm the sample still connects to the same peripheral.
+#. Change the address to a wrong value, rebuild and flash. Confirm the sample no
+   longer connects to any peripheral.
+#. Repeat the same flow with :kconfig:option:`CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_NAME`
+   (easier, uses the advertised name).
```
eivindj-nordic: I don't think we should keep this in the test steps, what do you think @anhmolt?

eivindj-nordic (approved):

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,133 @@
...
+* **Development**: Multiple HRS devices nearby, only want to connect to yours.
+* **Production**: Bridge paired with one specific sensor, ignore everything else.
+* **Bonded scenarios**: Combine with allow lists for tighter control.
+Note: "peripheral" in the option name refers to the *remote* device being scanned
+for, not this device's role.
+
+Testing
+=======
+
+This sample can be tested with three devices,
```
b-gent:
```suggestion

```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,133 @@
...
+This sample can be tested with three devices, one running this sample, another one running the :ref:`ble_hrs_sample` sample, and a central, for example, a phone or a tablet with `nRF Connect for Mobile`_ or `nRF Toolbox`_.
```
b-gent:
```suggestion
This sample can be tested with three devices:

* A device running this sample.
* A device running the :ref:`ble_hrs_sample` sample.
* A central device, for example, a phone or a tablet with `nRF Connect for Mobile`_ or `nRF Toolbox`_.
```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,133 @@
...
+The sample always scans for devices advertising the Heart Rate Service UUID
+(``0x180D``). Two optional filters can narrow this down further:
+
+.. list-table::
+   :header-rows: 1
+
+   * - Kconfig option
+     - Matches on
+   * - :kconfig:option:`CONFIG_SAMPLE_USE_TARGET_PERIPHERAL_NAME`
+     - Advertised device name (eks: ``"MyDeviceName"``)
```
b-gent:
```suggestion
     - Advertised device name (e.g. ``"MyDeviceName"``)
```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
b-gent:
```suggestion
If a central connects to this device, the central subscribes to the Heart Rate Measurement characteristic to receive heart rate notifications forwarded from the peripheral device.
```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,133 @@
...
+Building and running
+********************
+
+This sample can be found under :file:`samples/bluetooth/ble_hrs_peripheral_central/` in the |BMshort| folder structure.
+
+For details on how to create, configure, and program a sample, see :ref:`getting_started_with_the_samples`.
+
+Scan filtering options
+======================
+The sample always scans for devices advertising the Heart Rate Service UUID
```
b-gent:
```suggestion
The sample always scans for devices advertising the Heart Rate Service UUID (``0x180D``).
```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,133 @@
...
+(``0x180D``). Two optional filters can narrow this down further:
```
b-gent:
```suggestion
Two optional filters can narrow this down further:
```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
b-gent: I agree that the bullet points sound a bit like marketing content and don't really bring in new technical information. Please remove the bullets but retain the note, this is actually important.


```suggestion
.. note::

   ``peripheral`` in the option name refers to the *remote* device being scanned for, not this device's role.
```

samples/bluetooth/ble_hrs_peripheral_central/README.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Overview
+********
+
+This sample advertises and scans for devices that advertise with the :ref:`lib_ble_service_hrs` UUID (0x180D) and initiates a connection when a device is found.
+When a peripheral device is connected, the sample starts the service discovery procedure.
+If this succeeds, the sample subscribes to the Heart Rate Measurement characteristic to receive heart rate notifications.
+If a central connects to this device the central subscribes to the Heart Rate Measurement characteristic to receive heart rate notifications forwarded from the peripheral device.
+
+.. _ble_hrs_peripheral_central_sample_testing:
```
b-gent: this label should be moved before 'Building and running'

b-gent (changes requested):

samples/bluetooth/ble_hrs_peripheral_central/README.rst
anhmolt: Rewrote. Please have another look.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
anhmolt: Good catch. I see the same in the NUS central and HRS central sample READMEs. I'll fix in #799.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
anhmolt: Fixed

samples/bluetooth/ble_hrs_peripheral_central/README.rst
anhmolt: Fixed.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
anhmolt: Fixed.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
anhmolt: Removed bullet points. Kept the note.

samples/bluetooth/ble_hrs_peripheral_central/README.rst
anhmolt: fixed

samples/bluetooth/ble_hrs_peripheral_central/README.rst
anhmolt: Done

samples/bluetooth/ble_hrs_peripheral_central/README.rst
anhmolt: Removed this last step.

eivindj-nordic:

nordicjm (approved):

eivindj-nordic:
Dropping hrs_peripheral_central sample for now. Code fixes added to #829.

PR #697, opened 2026-03-17
Smaller clipboard's SRAM region & bloototh_name module removal + doc

nvlsianpu:
follow up to #633 : 
* Removed module of bluetooth name handling along with
zephyr-rtos: settings subsystem (CONFIG_NCS_BM_SETTINGS_BLUETOOTH_NAME).
It has been replaced by CONFIG_BM_FLAT_SETTINGS_BLUETOOTH_NAME.
* decrease size of Retained SRAM area in boards DTS'es  from 1 kB to 64 B
* update documentation on change in Bluetooth LE advertising name run-time setup
* minor documentation fixes

PR #730, opened 2026-04-07
Boards: Combine SoftDevice RAM into one node + update board layout images

anhmolt:
The nodes for SoftDevice allocated RAM is thus far not referenced in code. When modifying the allocated RAM for SoftDevice, there is no real need to distinguish between static and dynamic RAM. When getting numbers on the needed or required RAM
from the SoftDevice on initialization there is no distinction between static and dynamic.
The static RAM size is equal to the minimum size of the combined node.

Also update the board layout images in documentation after numerous changes to dts files.

### DIFF
--- a/boards/nordic/bm_nrf54l15dk/bm_nrf54l15dk_nrf54l05_cpuapp_s115_softdevice.dts
+++ b/boards/nordic/bm_nrf54l15dk/bm_nrf54l15dk_nrf54l05_cpuapp_s115_softdevice.dts
@@ -59,14 +59,9 @@
 &cpuapp_sram {
 	status = "okay";
 
-	softdevice_static_ram: sram@0 {
-		label = "softdevice_static_ram";
-		reg = <0x0 0x1080>;
-	};
-
-	softdevice_dynamic_ram: sram@1080 {
-		label = "softdevice_dynamic_ram";
-		reg = <0x1080 (DT_SIZE_K(12) + 0x300)>;
+	softdevice_ram: sram@0 {
+		label = "softdevice_ram";
+		reg = <0x0 0x4380>;
 	};
 
 	app_ram: sram@4380 {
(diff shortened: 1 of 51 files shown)


eivindj-nordic (approved):

boards/nordic/bm_nrf54l15dk/bm_nrf54l15dk_nrf54l05_cpuapp_s115_softdevice.dts
```diff
@@ -59,14 +59,9 @@
...
-		label = "softdevice_static_ram";
-		reg = <0x0 0x1080>;
-	};
-
-	softdevice_dynamic_ram: sram@1080 {
-		label = "softdevice_dynamic_ram";
-		reg = <0x1080 (DT_SIZE_K(12) + 0x300)>;
+	softdevice_ram: sram@0 {
+		label = "softdevice_ram";
+		reg = <0x0 (DT_SIZE_K(16) + 0x380)>;
```
lemrey: Would it be better to have it as an addition between the static and dynamic components?
For example `(0x1080 + 0x4380)`.

b-gent (approved):

nordicjm (approved):

boards/nordic/bm_nrf54l15dk/bm_nrf54l15dk_nrf54l05_cpuapp_s115_softdevice.dts
anhmolt: You have a valid point. The current way of specifying is not easier to read and neither is it any shorter. I will update.

eivindj-nordic:
DNM: Will be rebased after #697 is merged and images updated.

boards/nordic/bm_nrf54l15dk/bm_nrf54l15dk_nrf54l05_cpuapp_s115_softdevice.dts
lemrey: Alternatively, just put one number here.

anhmolt: Updated to use one number for the size.

anhmolt:
Regenerated images. Changed SoftDevice RAM color to match that of SoftDevice code partition.

PR #741, opened 2026-04-11
mgmt: mcumgr: img_mgmt: rewrite to claim data untill write completes

eivindj-nordic:
This allows only one write operation in flight at the same time, though the next will be started on the event of write complete if applicable.

Wait for all image write operations to complete before rebooting the firmware_loader application. This is currently using the new img_mgmt_write_in_progress() function directly in main, which works, but jumps over some layers of abstraction.


eivindj-nordic:
~Updates to changes done after previous release, no changelog required.~

Added new Kconfig options.

subsys/mgmt/mcumgr/grp/img_mgmt/src/bm_img_mgmt.c
```diff
@@ -107,63 +146,28 @@ int img_mgmt_read(int slot, unsigned int offset, void *dst, unsigned int num_byt
 int img_mgmt_write_image_data(unsigned int offset, const void *data, unsigned int num_bytes,
 			      bool last)
 {
-	int rc = IMG_MGMT_ERR_OK;
-	uint8_t *rb_data;
+	int rb_freespace;
```
anhmolt: free space is two words.
```suggestion
	int rb_free_space;
```
Or would it be clearer to call the variable `rb_unused`?

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
```diff
@@ -170,6 +170,25 @@ config MCUMGR_GRP_IMG_IMAGE_BUFFER_CHUNK_SZ
...
+	int "NVM write chunk size"
+	default 4
+	range 1 16
+	help
+	  This sets the maximum number of chunks that can be written to non-volatile memory
+	  in the same operation. The total size written to NVM should be small enough to clear the
+	  memory for new image data so that the operation does not time out.
+
+config MCUMGR_GRP_IMG_IMAGE_BUFFER_SZ
+	int "Image buffer size"
```
anhmolt, line 175: Would it make sense to let this be a hex value? Then the values would make more sense and look more sane.

I guess that would also impact `MCUMGR_GRP_IMG_IMAGE_BUFFER_CHUNK_SZ` and `MCUMGR_GRP_IMG_IMAGE_BUFFER_WRITE_CHUNKS_MAX` so maybe not such a great idea.
```suggestion
	hex "Image buffer size"
```

subsys/mgmt/mcumgr/grp/img_mgmt/src/bm_img_mgmt.c
eivindj-nordic: I think, `rb_space` might be enough.

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
eivindj-nordic: ~Yes, will do.~
Had to go back to int to set the range using MCUMGR_TRANSPORT_NETBUF_SIZE.

applications/firmware_loader/ble_mcumgr/src/main.c
```diff
@@ -169,6 +169,9 @@ static uint32_t ble_change_address(void)
 	return sd_ble_gap_addr_set(&device_address);
 }
 
+/* Forward declaration of extern function. */
+bool img_mgmt_write_in_progress(void);
```
nordicjm: put in header and drop this

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
```diff
@@ -163,13 +163,32 @@ config MCUMGR_GRP_IMG_SLOT_INFO_HOOKS
...
 
+config MCUMGR_GRP_IMG_IMAGE_BUFFER_WRITE_CHUNKS_MAX
+	int "NVM write chunk size"
+	default 4
+	range 1 16
+	help
+	  This sets the maximum number of chunks that can be written to non-volatile memory
+	  in the same operation. The total size written to NVM should be small enough to clear the
+	  memory for new image data so that the operation does not time out.
+
```
nordicjm: so I can configure `NVM write chunk size` and `NVM write chunk size` - what ?

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
```diff
@@ -163,13 +163,32 @@ config MCUMGR_GRP_IMG_SLOT_INFO_HOOKS
...
+config MCUMGR_GRP_IMG_IMAGE_BUFFER_SZ
+	hex "Image buffer size"
+	default 0x20000 if SOC_NRF54LM20A || SOC_NRF54L15 || SOC_NRF54L10 || SOC_NRF54LV10A
+	default 0xE000
+	range 0x10 0x20000
+	help
+	  This sets how much data can be received before written to non-volatile memory.
+	  It has to be larger or equal to CONFIG_MCUMGR_TRANSPORT_NETBUF_SIZE and
```
nordicjm: then use this as the start range value

nordicjm (changes requested):

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
eivindj-nordic: You can configure the chunk size, which is the size to write at minimum and a multiple of. And you can configure the maximum number of chunks to write in the same write operation. Grouping more chunks can improve performance, though we need to limit the maximum number of chunks to avoid scheduling the entire image buffer in the same write operation as that will take a long time and the BLE Device Manager app times out after too many retries of writing the same data fragment. 

Suggestions of better naming and description is welcome!

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
eivindj-nordic: Updated to use MCUMGR_TRANSPORT_NETBUF_SIZE as the start of the range. Had to rewrite to int prompt instead of hex.

applications/firmware_loader/ble_mcumgr/src/main.c
eivindj-nordic: Added to img_mgmt.h and moved that header to `include/bm/mgmt/...` to be exposed outside the library (as it is in zephyr).

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
nordicjm: I think you missed the point of both of these having the same prompt :)

eivindj-nordic: Yes, I did, thanks!

applications/firmware_loader/ble_mcumgr/src/main.c
eivindj-nordic: Actually, that did not work as `img_mgmt.h` has`#include <bootutil/image.h>`, which again require `mcuboot_config/mcuboot_config.h`, which is not available for the firmware loader image. Will leave header as is and add another one.

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
eivindj-nordic: Updated now!

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
```diff
@@ -162,13 +162,24 @@ config MCUMGR_GRP_IMG_SLOT_INFO_HOOKS
...
+	  clear the memory for new image data so that the operation does not time out.
+
+config MCUMGR_GRP_IMG_BUFFER_SZ
+	int "Image buffer size"
+	default 131072 if SOC_NRF54LM20A || SOC_NRF54L15 || SOC_NRF54L10 || SOC_NRF54LV10A
+	default 57344
+	range MCUMGR_TRANSPORT_NETBUF_SIZE 131072
+	help
+	  This sets how much data can be received before written to non-volatile memory.
+	  It has to be larger or equal to CONFIG_MCUMGR_TRANSPORT_NETBUF_SIZE and
```
peknis:
```suggestion
	  It has to be larger than or equal to CONFIG_MCUMGR_TRANSPORT_NETBUF_SIZE and
```

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
```diff
@@ -162,13 +162,24 @@ config MCUMGR_GRP_IMG_SLOT_INFO_HOOKS
...
+	  memory in the same operation. The total size written to NVM should be small enough to
+	  clear the memory for new image data so that the operation does not time out.
+
+config MCUMGR_GRP_IMG_BUFFER_SZ
+	int "Image buffer size"
+	default 131072 if SOC_NRF54LM20A || SOC_NRF54L15 || SOC_NRF54L10 || SOC_NRF54LV10A
+	default 57344
+	range MCUMGR_TRANSPORT_NETBUF_SIZE 131072
+	help
+	  This sets how much data can be received before written to non-volatile memory.
```
peknis:
```suggestion
	  This sets the volume of data that can be received before writing to non-volatile memory.
```

peknis (changes requested):

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
eivindj-nordic: Suggestion is not entirely correct, it is the volume of data that can be received and not yet written to NVM. Will rewrite.

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
```diff
@@ -162,13 +162,24 @@ config MCUMGR_GRP_IMG_SLOT_INFO_HOOKS
...
+	  memory in the same operation. The total size written to NVM should be small enough to
+	  clear the memory for new image data so that the operation does not time out.
+
+config MCUMGR_GRP_IMG_BUFFER_SZ
+	int "Image buffer size"
+	default 131072 if SOC_NRF54LM20A || SOC_NRF54L15 || SOC_NRF54L10 || SOC_NRF54LV10A
+	default 57344
+	range MCUMGR_TRANSPORT_NETBUF_SIZE 131072
+	help
+	  This sets the volume of data that can be received but not yet written to non-volatile
```
peknis:
```suggestion
	  This sets the volume of data that can be received but is not yet written to non-volatile
```

peknis (approved):

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
peknis: But the data is going to be written to non-volatile memory, right? Not sure how my suggestion was wrong.

eivindj-nordic: Yes, it is going to. Might be correct, though the way I interpret the suggested text is that this volume will be received before we begin writing, which is not correct. We start writing when we get at least the "wear unit" number of bytes.

subsys/mgmt/mcumgr/grp/img_mgmt/src/bm_img_mgmt.c
```diff
@@ -16,7 +17,7 @@
 #include <assert.h>
```
michalek-no, line 18: Asserts are not used anymore. Any chance to enforce buffer alignment in the build time?

eivindj-nordic: Will remove asserts include.
`bm_storage_nvm_info_get` require the instance to be initialized, so sadly it cannot be used at build time..

eivindj-nordic: (Or we can probably use (RRAMC_NRRAMWORDSIZE / BITS_PER_BYTE))

subsys/mgmt/mcumgr/grp/img_mgmt/src/bm_img_mgmt.c
```diff
@@ -107,63 +154,32 @@ int img_mgmt_read(int slot, unsigned int offset, void *dst, unsigned int num_byt
...
 
-	if ((rb_freespace - ongoing * CHUNK_SZ) < num_bytes) {
-		rc = MGMT_ERR_EBUSY;
+	if ((offset + num_bytes) > S0_SIZE) {
+		return IMG_MGMT_ERR_INVALID_IMAGE_TOO_LARGE;
+	}
+
+	rb_space = ring_buf_space_get(&ring_buf);
+	if (rb_space < num_bytes) {
+		LOG_DBG("busy");
```
michalek-no: already in img_mgmt.c

eivindj-nordic: Will remove the busy log.

eivindj-nordic: Done.

subsys/mgmt/mcumgr/grp/img_mgmt/src/bm_img_mgmt.c
eivindj-nordic: Added build time assert for SD and RRAM storage backends. Keeping assert header.

michalek-no (approved):

applications/firmware_loader/uart_mcumgr/src/main.c
lemrey, line 55: nit: Although we set this when we receive the reboot message, what this signals is rather that we have received the whole firmware. I would consider renaming this, e.g. `img_transfer_completed` or `img_transfer_in_progress`, so that the checks become a bit clearer
```
while (img_transfer_in_progress) { .. }
while (img_mgmt_write_in_progress()) { .. }
```

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
```diff
@@ -162,13 +162,24 @@ config MCUMGR_GRP_IMG_SLOT_INFO_HOOKS
...
 
-config MCUMGR_GRP_IMG_IMAGE_BUFFER_CHUNK_SZ
-	int "NVM write chunk size"
-	default 2048
-	range 16 4096
-	help
-	  This sets how much data is buffered before passing to non-volatile memory.
-	  It has to be a multiple of the NVM write block size.
+config MCUMGR_GRP_IMG_BUFFER_WRITE_WEAR_UNITS_MAX
+	int "NVM write max (in number of wear units)"
```
lemrey: I would like to find a better way to name this where we try to not mention wear-units. Here the user is tweaking `mcumgr` and they may not be familiar with that terminology (althought technically correct).

Perhaps this is simpler:
```
config MCUMGR_GRP_IMG_NVM_WRITE_BLOCKS_MAX
	int "Max NVM blocks per write operation"
	default 512
	range 1 2048
	help
	  Maximum number of NVM blocks (wear units) written to non-volatile memory
	  in a single operation. Keep this small enough so that each operation can
      complete quickly to free buffer for more incoming data.
```

subsys/mgmt/mcumgr/grp/img_mgmt/src/bm_img_mgmt.c
```diff
@@ -85,6 +86,46 @@ static const struct bm_storage_config s1_config = {
...
+	if ((ring_buf_size_get(&ring_buf) >= storage_wear_unit) || last_data) {
+		if (!atomic_cas(&ongoing, 0, 1)) {
+			/* Could not set ongoing, write opertion already in progress. */
+			return IMG_MGMT_ERR_OK;
+		}
+		rb_size = ring_buf_get_claim(&ring_buf, &rb_data, storage_write_size_max);
+
+		if (!last_data) {
+			/* Make sure we are writing full wear units until end of image. */
+			rb_size = (rb_size / storage_wear_unit) * storage_wear_unit;
```
lemrey: `zephyr/include/zephyr/sys/util.h` `ROUND_DOWN`

applications/firmware_loader/uart_mcumgr/src/main.c
```diff
@@ -54,6 +56,13 @@ int main(void)
...
 
 		smp_uart_process_rx_queue();
+		bm_scheduler_process();
+	}
+
+	/* Wait for the new firmware image to be written. */
+	while (img_mgmt_write_in_progress()) {
+		bm_scheduler_process();
+		k_sleep(K_MSEC(1));
 	}
```
lemrey, line 66: Is it possible to combine to two loops?

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
```diff
@@ -162,13 +162,24 @@ config MCUMGR_GRP_IMG_SLOT_INFO_HOOKS
...
+
+config MCUMGR_GRP_IMG_BUFFER_SZ
+	int "Image buffer size"
+	default 131072 if SOC_NRF54LM20A || SOC_NRF54L15 || SOC_NRF54L10 || SOC_NRF54LV10A
+	default 57344
+	range MCUMGR_TRANSPORT_NETBUF_SIZE 131072
+	help
+	  This sets the volume of data that can be received but is not yet written to non-volatile
+	  memory. It has to be larger than or equal to CONFIG_MCUMGR_TRANSPORT_NETBUF_SIZE and
+	  a multiple of the non-volatile memory wear unit.
```
lemrey, line 182: Why should this be a multiple of the wear unit? I don't think there is a requirement on the size of this buffer, only on the amount of bytes we ask `bm_storage` to write.

subsys/mgmt/mcumgr/grp/img_mgmt/src/bm_img_mgmt.c
```diff
@@ -29,36 +30,36 @@ LOG_MODULE_DECLARE(mcumgr_img_grp_data, CONFIG_MCUMGR_GRP_IMG_LOG_LEVEL);
...
-#define QUEUE_THRESHOLD (CONFIG_BM_STORAGE_BACKEND_SD_QUEUE_SIZE - WORST_CASE_TAIL_WRITES)
-BUILD_ASSERT(CONFIG_BM_STORAGE_BACKEND_SD_QUEUE_SIZE > WORST_CASE_TAIL_WRITES);
-BUILD_ASSERT(((PKTBUF_SZ % CHUNK_SZ) == 0) &&
-	(PKTBUF_SZ >= (CHUNK_SZ * CONFIG_BM_STORAGE_BACKEND_SD_QUEUE_SIZE)));
-BUILD_ASSERT((CHUNK_SZ % RRAMC_WRITE_BLOCK_SIZE) == 0);
-#else
-#define QUEUE_THRESHOLD 2
+#if defined(CONFIG_BM_STORAGE_BACKEND_SD) || defined(CONFIG_BM_STORAGE_BACKEND_RRAM)
+#define RRAMC_WEAR_UNIT (RRAMC_NRRAMWORDSIZE / BITS_PER_BYTE)
+BUILD_ASSERT((PKTBUF_SZ % RRAMC_WEAR_UNIT) == 0);
```
lemrey: I don't think this must be the case.

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
lemrey: Also, is 8k a good default?

scripts/ci/license_allow_list.yaml
```diff
@@ -34,37 +34,37 @@ any: |
...
+  ^nrf-bm/include/bm/mgmt/mcumgr/grp/img_mgmt/img_mgmt.h
   ^nrf-bm/lib/zephyr_queue/src/queue.c
   ^nrf-bm/modules/hal_nordic/nrfx/Kconfig
+  ^nrf-bm/subsys/fs/bm_zms/
   ^nrf-bm/subsys/mgmt/mcumgr/CMakeLists.txt
-  ^nrf-bm/subsys/mgmt/mcumgr/Kconfig
   ^nrf-bm/subsys/mgmt/mcumgr/grp/CMakeLists.txt
-  ^nrf-bm/subsys/mgmt/mcumgr/grp/Kconfig
   ^nrf-bm/subsys/mgmt/mcumgr/grp/img_mgmt/CMakeLists.txt
+  ^nrf-bm/subsys/mgmt/mcumgr/grp/img_mgmt/include/img_mgmt_priv.h
```
anhmolt: This path looks wrong.

applications/firmware_loader/uart_mcumgr/src/main.c
eivindj-nordic: Sure, though I don't think we should do that in this PR.

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
eivindj-nordic: Updated Kconfig with suggestion.

It is large enough to write multiple fragments received through BLE, and small enough so that the operation does not time out.

applications/firmware_loader/uart_mcumgr/src/main.c
eivindj-nordic: Probably, but lets take that in a separate PR.

subsys/mgmt/mcumgr/grp/img_mgmt/src/bm_img_mgmt.c
eivindj-nordic: See comment above. I think it makes sense to keep it though it is not strictly necessary now.

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
eivindj-nordic: It should be a multiple of the wear unit to reduce wear on the RRAM. If it is not a multiple it will retrieve a non-aligned number of bytes from the ring buffer (when reaching the end of the buffer). I've updated the code now to work in the case there is not write aligned bytes at the end of the ring buffer. This will allow non-aligned ring buffer size, though it is not optimal.

scripts/ci/license_allow_list.yaml
eivindj-nordic: Corrected

lemrey (approved):

anhmolt (approved):

eivindj-nordic:
@nordicjm Please have another look!

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
```diff
@@ -162,13 +162,24 @@ config MCUMGR_GRP_IMG_SLOT_INFO_HOOKS
...
 	  information to responses.
 
-config MCUMGR_GRP_IMG_IMAGE_BUFFER_CHUNK_SZ
-	int "NVM write chunk size"
-	default 2048
-	range 16 4096
-	help
-	  This sets how much data is buffered before passing to non-volatile memory.
-	  It has to be a multiple of the NVM write block size.
+config MCUMGR_GRP_IMG_NVM_WRITE_BLOCKS_MAX
```
nordicjm: since these are BM specific Kconfigs and not part of the original ones please add "BM_" to the Kconfigs like other things have, seems it was missed on the previous one that got removed

nordicjm (changes requested):

subsys/mgmt/mcumgr/grp/img_mgmt/Kconfig
anhmolt: Updated

nordicjm (approved):

PR #743, opened 2026-04-13
storage: bm_storage: send events async when possible

lemrey:
When the event callback fires from the same call context as the caller,
an application that queues another storage operation from inside its event
handler creates unbounded recursion. The chain looks like:

`bm_storage_write()` -> backend write -> `event_send()` -> user handler ->
`bm_storage_write()` -> backend write -> `event_send()` -> user handler -> ...

This applies to all backends.

Rewrite the `queue_process()` function in SoftDevice backend to be
iterative instead of recursive, to avoid unbounded recursion when
the SoftDevice is disabled.

This patch introduces a common function for backends that can be used
to defer the dispatching of the events from backends to the application
using the scheduler, when it is available. Otherwise, it just dispatches
the event immediately.

The other patch changes the behavior when the SoftDevice is busy and disabled,
where instead of draining the queue and possibly failing all operations,
we just stop and re-attempt to process what's left in the queue once
the user enqueues another operation.

### DIFF
--- a/doc/nrf-bm/libraries/bm_storage.rst
+++ b/doc/nrf-bm/libraries/bm_storage.rst
@@ -120,6 +120,18 @@ The following events may be reported to the user callback:
 
 Each event includes the result code, information about the address range of the associated operation, and whether the operation is synchronous or asynchronous.
 
+Event deferral
+--------------
+
+Backends that do not have an internal operation queue (for example, the RRAM backend) deliver events synchronously from the same call context as the caller.
+If the application queues another storage operation from within its event handler, unbounded recursion can occur.
+
+Backends with an internal operation queue (for example, the SoftDevice backend) are not affected, because re-entrant calls are enqueued and processed iteratively.
+
+To prevent recursion in backends without a queue, enable the :kconfig:option:`CONFIG_BM_SCHEDULER` Kconfig option.
+When the scheduler is available, synchronous events are automatically deferred and delivered from the main thread context on the next call to :c:func:`bm_scheduler_process`.
+Deferred events have their :c:member:`bm_storage_evt.is_async` field set to ``true``.
+
 Sample
 ******
 
@@ -141,6 +153,11 @@ Dependencies
   * :kconfig:option:`CONFIG_SOFTDEVICE`
   * :kconfig:option:`CONFIG_NRF_SDH`
   * :kconfig:option:`CONFIG_RING_BUFFER`
+
+Optional dependencies:
+
+* :kconfig:option:`CONFIG_BM_SCHEDULER` – Enables deferral of synchronous events to the main thread context.
+  See :ref:`Event deferral <lib_storage>`.
 
 API documentation
 *****************
--- a/doc/nrf-bm/release_notes/release_notes_changelog.rst
+++ b/doc/nrf-bm/release_notes/release_notes_changelog.rst
@@ -117,6 +117,7 @@ Storage
      * The :c:func:`bm_storage_nvm_info_get` function to retrieve NVM information, such as the size of the program unit and other.
      * The :c:member:`bm_storage_config.is_wear_aligned` configuration flag to enforce wear unit alignment on operations.
      * The :c:member:`bm_storage_config.is_write_padded` configuration flag to automatically pad write operations to the alignment unit.
+     * The :c:func:`bm_storage_evt_dispatch` function for event delivery.
 
    Updated:
      * The :c:func:`bm_storage_is_busy` function to return ``false`` instead of ``true`` when called with a ``NULL`` pointer or an uninitialized instance.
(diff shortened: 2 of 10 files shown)


subsys/storage/bm_storage/sd/bm_storage_sd.c
```diff
@@ -240,50 +242,62 @@ static void queue_process(void)
 {
 	uint32_t ret;
 
-	if (bm_storage_sd.operation_state == OP_NONE) {
-		if (!queue_load_next()) {
-			bm_storage_sd.queue_state = QUEUE_IDLE;
-			return;
+	for (;;) {
```
eivindj-nordic, line 245: while (true) ?

eivindj-nordic (approved):

subsys/storage/bm_storage/bm_storage.c
```diff
@@ -191,3 +195,45 @@ const struct bm_storage_info *bm_storage_nvm_info_get(const struct bm_storage *s
...
+#endif
+
+void bm_storage_evt_dispatch(const struct bm_storage *storage, struct bm_storage_evt *evt)
+{
+	if (!storage->evt_handler) {
+		return;
+	}
+
+#if defined(CONFIG_BM_SCHEDULER)
+	if (!evt->is_async) {
```
anhmolt, line 220: Maybe `IS_ENABLED()` can remove the need for all the `#if defined(CONFIG_BM_SCHEDULER) ... #endif`. Then this part of the code would always be checked by the compiler and discarded if `CONFIG_BM_SCHEDULER` is not set (`&&` operator short circuiting and compile time optimization). Not sure what is best.

```suggestion
	if (IS_ENABLED(CONFIG_BM_SCHEDULER) && !evt->is_async) {
```

doc/nrf-bm/libraries/bm_storage.rst
```diff
@@ -142,6 +154,10 @@ Dependencies
   * :kconfig:option:`CONFIG_NRF_SDH`
   * :kconfig:option:`CONFIG_RING_BUFFER`
 
+Optional dependencies:
+
+* :kconfig:option:`CONFIG_BM_SCHEDULER` – Enables deferral of synchronous events to the main thread context. See :ref:`Event deferral <lib_storage>`.
```
peknis:
```suggestion
* :kconfig:option:`CONFIG_BM_SCHEDULER` – Enables deferral of synchronous events to the main thread context.
  See :ref:`Event deferral <lib_storage>`.
```

doc/nrf-bm/release_notes/release_notes_changelog.rst
```suggestion
     * The ``start_addr`` and ``end_addr`` fields in :c:struct:`bm_storage_config` and :c:struct:`bm_storage` have been replaced by ``addr`` and ``size``.
       The API now uses relative addressing (0-based offsets within the partition).
```

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -131,6 +131,8 @@ Storage
      * The :c:func:`bm_storage_write` and :c:func:`bm_storage_erase` functions to return ``-ENOMEM`` when out of memory, instead of ``-EIO``.
      * The :c:func:`bm_storage_read`, :c:func:`bm_storage_write`, and :c:func:`bm_storage_erase` functions to return ``-EINVAL`` on alignment errors, instead of ``-EFAULT``.
      * The :c:enum:`bm_storage_evt_dispatch_type` enum and the :c:member:`bm_storage_evt.dispatch_type` field have been replaced by a boolean :c:member:`bm_storage_evt.is_async`.
+     * All backends to use the new :c:func:`bm_storage_evt_dispatch` function for event delivery. For backends without an internal operation queue (such as RRAM), enabling :c:func:`bm_scheduler` defers synchronous events to the main thread context.
+     * The SoftDevice backend to process operations iteratively instead of recursively when the SoftDevice is disabled. Re-entrant calls from event handlers are safely enqueued and processed by the loop.
```
```suggestion
     * All backends to use the new :c:func:`bm_storage_evt_dispatch` function for event delivery.
       For backends without an internal operation queue (such as RRAM), enabling :c:func:`bm_scheduler` defers synchronous events to the main thread context.
     * The SoftDevice backend to process operations iteratively instead of recursively when the SoftDevice is disabled.
       Re-entrant calls from event handlers are safely enqueued and processed by the loop.
```

peknis (changes requested):

anhmolt (approved):

subsys/storage/bm_storage/bm_storage.c
lemrey: That wouldn't work in this case because `bm_scheduler.h` and `deferred_evt_handler` are included/declared conditionally so the code wouldn't compile when `CONFIG_BM_SCHEDULER` is not enabled.

doc/nrf-bm/libraries/bm_storage.rst
lemrey: Applied

doc/nrf-bm/release_notes/release_notes_changelog.rst
lemrey: Applied

doc/nrf-bm/release_notes/release_notes_changelog.rst
lemrey: Applied

peknis (approved):

PR #747, opened 2026-04-14
tests: add sdh_evt_dispatch helper file

lemrey:
Add a small helper file to dispatch SoftDevice events to registered observers. This avoids boilerplate in tests, and clutter in the implementations under test to keep the observer's handlers global when compiling unit test.

### DIFF
--- a/cmake/unity/unity_softdevice_setup.cmake
+++ b/cmake/unity/unity_softdevice_setup.cmake
@@ -23,3 +23,18 @@ function(unity_softdevice_header_setup)
   zephyr_include_directories(${SOFTDEVICE_INCLUDE_DIR})
   zephyr_compile_definitions(SVCALL_AS_NORMAL_FUNCTION)
 endfunction()
+
+# Add the sdh_evt_dispatch test helper to the current unit test.
+#
+# The helper exposes sdh_evt_dispatch_{ble,soc,state}() which let tests
+# fire SoftDevice events at all observers registered via
+# NRF_SDH_BLE_OBSERVER, NRF_SDH_SOC_OBSERVER, and NRF_SDH_STATE_EVT_OBSERVER.
+#
+
+function(unity_softdevice_event_setup)
+  set(dir ${ZEPHYR_NRF_BM_MODULE_DIR}/tests/unit/util/sdh_evt_dispatch)
+
+  target_sources(app PRIVATE ${dir}/sdh_evt_dispatch.c)
+  target_include_directories(app PRIVATE ${dir}/include)
+  zephyr_linker_sources(SECTIONS ${dir}/sdh_evt_dispatch.ld)
+endfunction()
--- a/cmake/unity/unity_softdevice_setup.cmake
+++ b/cmake/unity/unity_softdevice_setup.cmake
@@ -1,0 +1,25 @@
+#
+# Copyright (c) 2026 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+function(unity_softdevice_header_setup)
+  cmake_parse_arguments(ARG "" "VARIANT;SOC" "" ${ARGN})
+
+  if(NOT DEFINED ARG_VARIANT)
+    set(ARG_VARIANT "s115")
+  endif()
+  if(NOT DEFINED ARG_SOC)
+    set(ARG_SOC "nrf54l")
+  endif()
+
+  set(SOFTDEVICE_INCLUDE_DIR
+    "${ZEPHYR_NRF_BM_MODULE_DIR}/components/softdevice/${ARG_SOC}/${ARG_VARIANT}/${ARG_VARIANT}_API/include"
+  )
+
+  set(SOFTDEVICE_INCLUDE_DIR "${SOFTDEVICE_INCLUDE_DIR}" PARENT_SCOPE)
+
+  zephyr_include_directories(${SOFTDEVICE_INCLUDE_DIR})
+  zephyr_compile_definitions(SVCALL_AS_NORMAL_FUNCTION)
+endfunction()
--- a/include/bm/bluetooth/ble_adv.h
+++ b/include/bm/bluetooth/ble_adv.h
@@ -238,10 +238,17 @@ struct ble_adv_config {
 };
 
 /**
- * @brief Library's Bluetooth LE event handler.
+ * @brief Bluetooth LE event handler for the advertising library.
+ *
+ * @details Handles all Bluetooth LE stack events that are of interest to the
+ *          advertising library.
+ *
+ * @note This handler is registered automatically by @ref BLE_ADV_DEF and is
+ *       called by the SoftDevice handler. The application does not need to call
+ *       it directly.
  *
  * @param[in] ble_evt Bluetooth LE stack event.
- * @param[in] ble_adv Advertising Module instance.
+ * @param[in] ble_adv Pointer to the @ref ble_adv instance.
  */
 void ble_adv_on_ble_evt(const ble_evt_t *ble_evt, void *ble_adv);
 
--- a/include/bm/bluetooth/ble_gq.h
+++ b/include/bm/bluetooth/ble_gq.h
@@ -300,16 +300,19 @@ uint32_t ble_gq_item_add(const struct ble_gq *gatt_queue, struct ble_gq_req *req
 uint32_t ble_gq_conn_handle_register(const struct ble_gq *gatt_queue, uint16_t conn_handle);
 
 /**
- * @brief Handle Bluetooth LE events from the SoftDevice.
+ * @brief Bluetooth LE event handler for the GATT Queue library.
  *
- * @details This function handles the Bluetooth LE events received from the SoftDevice. If an
- *          event is relevant to the BGQ module, it is used to update internal variables,
- *          process queued GATT requests and, if necessary, send errors to the application.
+ * @details Handles all Bluetooth LE stack events that are of interest to the
+ *          GATT Queue library.
  *
- * @param[in] ble_evt     Pointer to the event.
- * @param[in] gatt_queue  Pointer to the @ref ble_gq instance.
+ * @note This handler is registered automatically by @ref BLE_GQ_DEF and is
+ *       called by the SoftDevice handler. The application does not need to call
+ *       it directly.
+ *
+ * @param[in] ble_evt Bluetooth LE stack event.
+ * @param[in] ble_gq  Pointer to the @ref ble_gq instance.
  */
-void ble_gq_on_ble_evt(const ble_evt_t *ble_evt, void *gatt_queue);
+void ble_gq_on_ble_evt(const ble_evt_t *ble_evt, void *ble_gq);
 
 #ifdef __cplusplus
 }
--- a/include/bm/bluetooth/ble_qwr.h
+++ b/include/bm/bluetooth/ble_qwr.h
@@ -13,9 +13,6 @@
  *
  * @details This module handles prepare write, execute write, and cancel write
  * commands. It also manages memory requests related to these operations.
- *
- * @note The application must propagate Bluetooth LE stack events to this module by calling
- *       @ref ble_qwr_on_ble_evt().
  */
 
 #ifndef BLE_QWR_H__
@@ -180,15 +177,19 @@ uint32_t ble_qwr_init(struct ble_qwr *qwr, const struct ble_qwr_config *qwr_conf
 uint32_t ble_qwr_conn_handle_assign(struct ble_qwr *qwr, uint16_t conn_handle);
 
 /**
- * @brief Function for handling Bluetooth LE stack events.
+ * @brief Bluetooth LE event handler for the Queued Writes module.
  *
- * @details Handles all events from the Bluetooth LE stack that are of interest to the
+ * @details Handles all Bluetooth LE stack events that are of interest to the
  *          Queued Writes module.
  *
- * @param[in] ble_evt Event received.
- * @param[in] context Queued Writes structure.
+ * @note This handler is registered automatically by @ref BLE_QWR_DEF and is
+ *       called by the SoftDevice handler. The application does not need to call
+ *       it directly.
+ *
+ * @param[in] ble_evt Bluetooth LE stack event.
+ * @param[in] ble_qwr Pointer to the @ref ble_qwr instance.
  */
-void ble_qwr_on_ble_evt(const ble_evt_t *ble_evt, void *context);
+void ble_qwr_on_ble_evt(const ble_evt_t *ble_evt, void *ble_qwr);
 
 #if (CONFIG_BLE_QWR_MAX_ATTR > 0)
 /**
--- a/tests/unit/lib/bluetooth/ble_adv/CMakeLists.txt
+++ b/tests/unit/lib/bluetooth/ble_adv/CMakeLists.txt
@@ -25,7 +25,6 @@ zephyr_compile_definitions(
 
 zephyr_include_directories(
   ${SOFTDEVICE_INCLUDE_DIR}
-  ${ZEPHYR_NRF_BM_MODULE_DIR}/include
 )
 
 # Generate and add test file
--- a/tests/unit/lib/bluetooth/ble_adv/CMakeLists.txt
+++ b/tests/unit/lib/bluetooth/ble_adv/CMakeLists.txt
@@ -16,11 +16,6 @@ unity_softdevice_header_setup()
 cmock_handle(${SOFTDEVICE_INCLUDE_DIR}/ble.h)
 cmock_handle(${SOFTDEVICE_INCLUDE_DIR}/ble_gap.h)
 
-zephyr_compile_definitions(
-  NRF54L15_XXAA
-  SUPPRESS_INLINE_IMPLEMENTATION
-)
-
 # Generate and add test file
 test_runner_generate(src/unity_test.c)
 target_sources(app PRIVATE src/unity_test.c)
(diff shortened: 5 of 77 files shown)


tests/common/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+#
+# Copyright (c) 2026 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+add_subdirectory_ifdef(CONFIG_UNITY sdh_evt_dispatch)
```
eivindj-nordic: I think we should have a separate Kconfig for including this.

lemrey: I thought about it, but I wasn't sure. Of course it wouldn't hurt, but it doesn't hurt either to have it available by default when writing unit tests, and that's one KConfig less.

subsys/softdevice/Kconfig
```diff
@@ -7,8 +7,8 @@ menu "SoftDevice"
 
 config SOFTDEVICE
 	bool "SoftDevice is present"
-	depends on ZERO_LATENCY_IRQS
-	imply NRF_SDH
+	depends on (ZERO_LATENCY_IRQS || UNITY)
+	imply NRF_SDH if !UNITY
```
eivindj-nordic: I'd rather keep the include paths as is in the unit tests CMake than add UNITY Kconfig dependencies here.  We're sort of adding complexity to the code base that is only relevant for our tests.

tests/common/CMakeLists.txt
eivindj-nordic: Dependency on SoftDevice Kconfig as it is now is fine.

subsys/softdevice/Kconfig
lemrey: I see your point, it's not wrong.
Let me see if I can adjust that.

lemrey: I made a CMake function instead

eivindj-nordic (approved):
There are places where we call the BLE event handler directly, e.g. `ble_bas_client_on_ble_evt`. Not sure if it is desired to rewrite to use `sdh_evt_dispatch_ble` etc. also here(?).

lemrey:
> There are places where we call the BLE event handler directly, e.g. ble_bas_client_on_ble_evt. Not sure if it is desired to rewrite to use sdh_evt_dispatch_ble etc. also here(?).

Yes, thank you. I must have missed it

lemrey:
Actually, I have to think about that. It may be fine to keep those, and in fact a bit easier so that each test can keep defining its instance rather than have a common instance that is shared by all tests (using `_DEF` macros), that then needs to be reset every time.

eivindj-nordic:
> Actually, I have to think about that. It may be fine to keep those, and in fact a bit easier so that each test can keep defining its instance rather than have a common instance that is shared by all tests (using _DEF macros), that then needs to be reset every time.

With the remaining tests, the ble event function is exposed in the public header, so it is fine to keep as is as well.

lemrey:
> With the remaining tests, the ble event function is exposed in the public header, so it is fine to keep as is as well.

I agree, I made sure we use the `sdh_evt_dispatch` only for those tests that have `static` event handlers.
The rest invoke the public function instead. I added one commit to align those i.e. doc, basic error checking, param naming, visibility.

include/bm/bluetooth/ble_gq.h
```diff
@@ -300,16 +300,19 @@ uint32_t ble_gq_item_add(const struct ble_gq *gatt_queue, struct ble_gq_req *req
...
- * @param[in] gatt_queue  Pointer to the @ref ble_gq instance.
+ * @note This handler is registered automatically by @ref BLE_GQ_DEF and is
+ *       called by the SoftDevice handler. The application does not need to call
+ *       it directly.
+ *
+ * @param[in] ble_evt Bluetooth LE stack event.
+ * @param[in] gq      Pointer to the @ref ble_gq instance.
  */
-void ble_gq_on_ble_evt(const ble_evt_t *ble_evt, void *gatt_queue);
+void ble_gq_on_ble_evt(const ble_evt_t *ble_evt, void *gq);
```
anhmolt: What is the reason for changing the parameter name here? The other function declarations in this file uses `gatt_queue`. I guess using `gq` aligns better with the naming scheme of other libraries.

include/bm/bluetooth/ble_adv.h
```diff
@@ -238,12 +238,19 @@ struct ble_adv_config {
...
+ * @note This handler is registered automatically by @ref BLE_ADV_DEF and is
+ *       called by the SoftDevice handler. The application does not need to call
+ *       it directly.
  *
  * @param[in] ble_evt Bluetooth LE stack event.
- * @param[in] ble_adv Advertising Module instance.
+ * @param[in] adv     Pointer to the @ref ble_adv instance.
  */
-void ble_adv_on_ble_evt(const ble_evt_t *ble_evt, void *ble_adv);
+void ble_adv_on_ble_evt(const ble_evt_t *ble_evt, void *adv);
```
anhmolt: Rest of this file (except the adv event handler) uses `ble_adv` for the instance parameter name. Should we stick to that here?

cmake/unity/unity_softdevice_setup.cmake
```diff
@@ -0,0 +1,25 @@
...
+
+  if(NOT ARG_VARIANT)
+    set(ARG_VARIANT "s115")
+  endif()
+  if(NOT ARG_SOC)
+    set(ARG_SOC "nrf54l")
+  endif()
+
+  set(SOFTDEVICE_INCLUDE_DIR
+    "${ZEPHYR_NRF_BM_MODULE_DIR}/components/softdevice/${ARG_SOC}/${ARG_VARIANT}/${ARG_VARIANT}_API/include"
```
anhmolt, line 18: Would it make sense to store the path to a local variable first and then reuse that in both the `set(... PARENT_SCOPE)` and the `zephyr_include_directories`?

tests/unit/cmake/softdevice_unity_setup.cmake
```diff
@@ -0,0 +1,25 @@
...
+  if(NOT ARG_VARIANT)
+    set(ARG_VARIANT "s115")
+  endif()
+  if(NOT ARG_SOC)
+    set(ARG_SOC "nrf54l")
+  endif()
+
+  set(SOFTDEVICE_INCLUDE_DIR
+    "${ZEPHYR_NRF_BM_MODULE_DIR}/components/softdevice/${ARG_SOC}/${ARG_VARIANT}/${ARG_VARIANT}_API/include"
+    PARENT_SCOPE)
```
anhmolt: Nit. Consider placing the ending parenthesis on the next line. Looking through sdk-nrf, that seems like the prevalent CMake codestyle for functions spanning multiple lines.

tests/unit/cmake/softdevice_unity_setup.cmake
```diff
@@ -0,0 +1,25 @@
...
+
+  zephyr_include_directories(
+    "${ZEPHYR_NRF_BM_MODULE_DIR}/components/softdevice/${ARG_SOC}/${ARG_VARIANT}/${ARG_VARIANT}_API/include")
```
anhmolt:
```suggestion
    "${ZEPHYR_NRF_BM_MODULE_DIR}/components/softdevice/${ARG_SOC}/${ARG_VARIANT}/${ARG_VARIANT}_API/include"
  )
```

tests/unit/cmake/softdevice_unity_setup.cmake
```diff
@@ -0,0 +1,25 @@
...
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+function(softdevice_unity_setup)
+  cmake_parse_arguments(ARG "" "VARIANT;SOC" "" ${ARGN})
+
+  if(NOT ARG_VARIANT)
+    set(ARG_VARIANT "s115")
+  endif()
+  if(NOT ARG_SOC)
```
anhmolt:
```suggestion
  if(NOT DEFINED ARG_VARIANT)
    set(ARG_VARIANT "s115")
  endif()
  if(NOT DEFINED ARG_SOC)
```
Variable existence vs. logical evaluation on the variable.
It's not so important here since you would "never" pass `OFF`, `0`, `FALSE`, `NO`, etc. as the SoftDevice variant or SoC. But if you do pass `off` as variant, it will set `ARG_VARIANT` to `s115`.

tests/unit/lib/bluetooth/ble_db_discovery/CMakeLists.txt
```diff
@@ -10,23 +10,10 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 
 project(unit_test_ble_db_discovery)
 
-set(SOFTDEVICE_VARIANT "s145")
-set(SOFTDEVICE_INCLUDE_DIR "${ZEPHYR_NRF_BM_MODULE_DIR}/components/softdevice/nrf54l/\
-${SOFTDEVICE_VARIANT}/${SOFTDEVICE_VARIANT}_API/include")
+softdevice_unity_setup(VARIANT s145)
```
anhmolt:
```suggestion
softdevice_unity_setup(VARIANT "s145")
```
Not important, but should we quote these strings? [Quote Strings](https://docs.zephyrproject.org/latest/contribute/style/cmake.html#quoting)

lemrey:
@nrfconnect/ncs-co-build-system , @nrfconnect/ncs-code-owners please review

tests/unit/subsys/storage/bm_storage_sd/src/unity_test.c
```diff
@@ -28,11 +28,11 @@
 #define WORD_SIZE(a) ((a) / sizeof(uint32_t))
 #define PTR_IGNORE NULL
 
-/* The backend's SoC event handler */
-extern void bm_storage_sd_on_soc_evt(uint32_t evt, void *ctx);
 /* The backend's SoftDevice state event handler */
 extern int bm_storage_sd_on_state_evt(enum nrf_sdh_state_evt evt, void *ctx);
```
anhmolt: Any reason why `bm_storage_sd_on_state_evt` is still used in the test and not swapped with `sdh_evt_dispatch_state`?

Should the `static` keyword removal when `CONFIG_UNITY` is set also be removed from https://github.com/nrfconnect/sdk-nrf-bm/blob/3b39162ae623147e84abec18c78033c08c397f2e/subsys/storage/bm_storage/sd/bm_storage_sd.c#L462-L465
Similar to how it has been done for `bm_storage_sd_on_soc_evt`, or is there a reason why this have been kept?

include/bm/bluetooth/ble_qwr.h
```diff
@@ -180,15 +180,19 @@ uint32_t ble_qwr_init(struct ble_qwr *qwr, const struct ble_qwr_config *qwr_conf
...
  *
- * @details Handles all events from the Bluetooth LE stack that are of interest to the
+ * @details Handles all Bluetooth LE stack events that are of interest to the
  *          Queued Writes module.
  *
- * @param[in] ble_evt Event received.
- * @param[in] context Queued Writes structure.
+ * @note This handler is registered automatically by @ref BLE_QWR_DEF and is
+ *       called by the SoftDevice handler. The application does not need to call
+ *       it directly.
```
anhmolt, line 187: This points in a slightly different direction than a note at the start of the file: https://github.com/nrfconnect/sdk-nrf-bm/blob/3b39162ae623147e84abec18c78033c08c397f2e/include/bm/bluetooth/ble_qwr.h#L17-L18

I suggest removing the note at the top of the file (this is not mentioned in any similar header).

anhmolt (commented):
Good cleanup! See suggestions/comments.

include/bm/bluetooth/ble_adv.h
lemrey: It seems that `ble_adv` and `ble_gq` are outliers, most libraries omit the `ble_` part in the name of the instance parameter.

include/bm/bluetooth/ble_gq.h
lemrey: > I guess using gq aligns better with the naming scheme of other libraries.

Yes, see [my other comment](https://github.com/nrfconnect/sdk-nrf-bm/pull/747#discussion_r3166915468)

include/bm/bluetooth/ble_qwr.h
lemrey: Good catch

include/bm/bluetooth/ble_adv.h
lemrey: But this made me think it's best to flip the approach. Instead of naming the `void *` param just `adv` or `hrs` we can specify the full datatype name e.g. `ble_adv` and then in the C file implementation cast it to its right type into a variable that has the same name as the instance in all the other functions.

cmake/unity/unity_softdevice_setup.cmake
lemrey: Ok, why not

tests/unit/cmake/softdevice_unity_setup.cmake
lemrey: Ok

tests/unit/cmake/softdevice_unity_setup.cmake
lemrey: Updated

tests/unit/lib/bluetooth/ble_db_discovery/CMakeLists.txt
lemrey: Good point

tests/unit/subsys/storage/bm_storage_sd/src/unity_test.c
lemrey: Right, I missed this; we should do it here too.

anhmolt (approved):

tests/unit/cmake/softdevice_unity_setup.cmake
lemrey: Updated

tests/unit/util/sdh_evt_dispatch/Kconfig
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) 2026 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config UNITY_SDH_EVT_DISPATCH
```
nordicjm: why are we sourcing test Kconfigs globally which have no use for a customer, and shouldn't even be in the Kconfig tree ?

lemrey: There is a C file in `/test` which I'd like to compile and make available in unit tests, so I made a Kconfig for that and I am sourcing it. If there are different ways of approaching this, could you please suggest what they are so that I can update the patch?

tests/unit/subsys/bluetooth/services/ble_hrs_client/CMakeLists.txt
```diff
@@ -10,25 +10,14 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 
 project(unit_test_ble_hrs_client)
 
-set(SOFTDEVICE_VARIANT "s115")
-set(SOFTDEVICE_INCLUDE_DIR "${ZEPHYR_NRF_BM_MODULE_DIR}/components/softdevice/nrf54l/\
-${SOFTDEVICE_VARIANT}/${SOFTDEVICE_VARIANT}_API/include")
+softdevice_unity_setup()
```
anhmolt: Think it makes sense to change to S145 here.
```suggestion
softdevice_unity_setup(VARIANT "s145")
```

lemrey: Updated

tests/unit/util/sdh_evt_dispatch/Kconfig
lemrey: @nordicjm do you have any input here?

nordicjm: No sourcing of files in test globally, use a relative include or whatnot on tests that need it

lemrey: Does it look better now @nordicjm ?

CMakeLists.txt
```diff
@@ -12,4 +12,5 @@ if(CONFIG_NCS_BM)
   add_subdirectory(drivers)
   add_subdirectory(lib)
   add_subdirectory(subsys)
+  add_subdirectory(tests)
```
nordicjm: not better no, again, tests should not be part of the cmake/Kconfig tree

nordicjm (changes requested):

CMakeLists.txt
lemrey: I see, I gave it another try @nordicjm

marenges (approved):

eivindj-nordic:
@nordicjm Please have another look.

tests/unit/lib/bluetooth/ble_adv/CMakeLists.txt
```diff
@@ -10,23 +10,16 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 
 project(unit_test_ble_adv)
 
-set(SOFTDEVICE_VARIANT "s115")
-set(SOFTDEVICE_INCLUDE_DIR "${ZEPHYR_NRF_BM_MODULE_DIR}/components/softdevice/nrf54l/\
-${SOFTDEVICE_VARIANT}/${SOFTDEVICE_VARIANT}_API/include")
+softdevice_unity_setup()
```
nordicjm: why can't these files do `include(${ZEPHYR_NRF_BM_MODULE_DIR}/cmake/unity/softdevice_unity_setup.cmake)` ?

lemrey: Updated!

lemrey: @nordicjm please have a look when you can

nordicjm (approved):

PR #755, opened 2026-04-20
samples: bluetooth: move initialized log before adv_start

eivindj-nordic:
Advertising start is not part of the initialization. This also matches the expected output in the documentation.

### DIFF
--- a/samples/bluetooth/ble_pwr_profiling/src/main.c
+++ b/samples/bluetooth/ble_pwr_profiling/src/main.c
@@ -703,7 +703,11 @@ int main(void)
 		goto idle;
 	}
 
-	LOG_INF("Services initialized");
+#if defined(CONFIG_SAMPLE_BLE_PWR_PROFILING_LED)
+	nrf_gpio_pin_write(BOARD_PIN_LED_0, BOARD_LED_ACTIVE_STATE);
+#endif
+
+	LOG_INF("BLE PWR Profiling sample initialized");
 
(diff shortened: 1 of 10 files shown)


lemrey:
I think we intended it to be printed right before entering the `main` loop, am I wrong?

eivindj-nordic:
> I think we intended it to be printed right before entering the main loop, am I wrong?

We did, though looking at it with documentation, and the fact that it says initialized, not started (which we have when the sample starts and we start to initialize, it does not really make sense. 

Now the documentation says e.g. "observe sample initialized" in terminal, then "observe sample start advertising" in terminal. Though, these are logged in opposite order. It would not make sense to rewrite the documentation to observe the sample to start advertising before observing the sample to be initialized, I think.

lemrey (approved):
That's fine

samples/bluetooth/ble_pwr_profiling/src/main.c
```diff
@@ -703,7 +703,11 @@ int main(void)
 		goto idle;
 	}
 
-	LOG_INF("Services initialized");
+	#if defined(CONFIG_SAMPLE_BLE_PWR_PROFILING_LED)
```
anhmolt:
```suggestion
#if defined(CONFIG_SAMPLE_BLE_PWR_PROFILING_LED)
```

anhmolt (approved):
Approving with one nit.

PR #761, opened 2026-04-21
samples: SPI sample

PizzaAllTheWay:
Add SPI sample with SPIM and SPIS support for single-device loopback and two-device setups (one or two data paths). Button 2 triggers a master transfer, LED 2 toggles on slave reception.

Includes:
- SPIM and SPIS initialization and transfer handling
- Board configuration with default pin mappings
- Kconfig option for configurable message string
- Documentation with wiring guide, pin conflict notes, and testing steps for all three wiring configurations


boards/nordic/bm_nrf54l15dk/include/board-config.h
```diff
@@ -63,58 +63,79 @@ extern "C" {
...
+#define BOARD_APP_UARTE_PIN_TX NRF_PIN_PORT_TO_PIN_NUMBER(0, 0) /* Shared with SPI slave SCK. */
+#endif
+#ifndef BOARD_APP_UARTE_PIN_RX
+#define BOARD_APP_UARTE_PIN_RX NRF_PIN_PORT_TO_PIN_NUMBER(1, 0) /* Shared with SPI slave MOSI. */
+#endif
+#ifndef BOARD_APP_UARTE_PIN_RTS
+#define BOARD_APP_UARTE_PIN_RTS NRF_PIN_PORT_TO_PIN_NUMBER(2, 0) /* Shared with SPI slave MISO. */
+#endif
+#ifndef BOARD_APP_UARTE_PIN_CTS
+#define BOARD_APP_UARTE_PIN_CTS NRF_PIN_PORT_TO_PIN_NUMBER(3, 0) /* Shared with SPI slave CSN. */
```
eivindj-nordic: Let's not add these comments here. If any it should be the other way around.

boards/nordic/bm_nrf54l15dk/include/board-config.h
```diff
@@ -63,58 +63,79 @@ extern "C" {
 #define BOARD_CONSOLE_UARTE_PIN_CTS NRF_PIN_PORT_TO_PIN_NUMBER(7, 1)
 #endif
 
-/* Application UART configuration */
-#ifndef BOARD_APP_UARTE_INST
-#define BOARD_APP_UARTE_INST NRF_UARTE30
+/* Application SPI master configuration */
+#ifndef BOARD_APP_SPIM_INST
```
eivindj-nordic, line 67: We need board config for the other boards as well.

samples/peripherals/spi/CMakeLists.txt
```diff
@@ -1,12 +1,12 @@
 #
-# Copyright (c) 2026 Nordic Semiconductor ASA
+# Copyright (c) 2025 Nordic Semiconductor ASA
```
eivindj-nordic:
```suggestion
# Copyright (c) 2026 Nordic Semiconductor ASA
```

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+
+      =================  =================
+      Device 1 (SPIM)    Device 2 (SPIS)
+      =================  =================
+      SCK                SCK
+      MOSI               MOSI
+      MISO               MISO
+      CSN                CSN
+      **GND**            **GND**
+      =================  =================
```
eivindj-nordic: Use same formatting as other tables

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+The SPI sample demonstrates how to configure and use the SPIM and SPIS peripherals with the nrfx drivers.
+
+Requirements
+************
+
+The sample supports the following development kits:
+
+.. tabs::
+
+   .. group-tab:: Simple board variants
```
eivindj-nordic, line 19: Use templates for this

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+           - S145
+           - bm_nrf54ls05dk/nrf54ls05b/cpuapp/s145_softdevice/mcuboot
+         * - `nRF54LV10 DK`_
+           - PCA10188
+           - S145
+           - bm_nrf54lv10dk/nrf54lv10a/cpuapp/s145_softdevice/mcuboot
+
+.. _bm_spi_wiring:
+
+Wiring
```
eivindj-nordic: This belongs under test step

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+         * - MISO
+           - MISO
+         * - CSN
+           - CSN
+
+      Connect **GND** between the two devices.
+
+      Press **Button 2** on either device to send to the other.
+
+.. note:: LED 3 and SPIM CSN (nRF54L15)
```
eivindj-nordic: We should not have four notes after each other.

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+.. note:: Button 0 and SPIM MISO (nRF54L15)
+
+   On the `nRF54L15 DK`_, P1.13 is shared between **SPIM MISO** and **Button 0**.
+   Do not press Button 0 while the SPI bus is active.
+
+.. note:: UARTE pin conflict (nRF54L15)
+
+   On the `nRF54L15 DK`_, **SPIS** pins P0.00-P0.03 overlap with the default **Application UARTE** in :file:`board-config.h`.
+   The :ref:`uarte_sample` and this sample cannot use those pins at the same time.
+   Reassign one peripheral in the board file if both are needed.
```
eivindj-nordic: This can be removed. The sample does not use application uarte.

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+
+.. note:: LPUARTE pin conflict (nRF54L15)
+
+   On the `nRF54L15 DK`_, **LPUARTE TX, RX, and RDY** share P1.11, P1.12, and P1.14 with **SPIM SCK, MOSI, and CSN**. **LPUARTE REQ** (P0.04) overlaps with **Button 3**.
+   The :ref:`bm_lpuarte_sample` and this sample cannot run together without reassigning pins in the board file.
```
eivindj-nordic: We run one sample at a time. This comment is not needed.

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+
+Optionally, connect a logic analyzer or oscilloscope to the SPI lines.
+Use a short GND lead to the dev kit ground and a sample rate high enough to resolve the 4 MHz default SPI clock.
```
eivindj-nordic: This is part of testing and if kept should be placed there. I think details on testing with logic analyzer is something we can remove.

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+
+Overview
+********
+
+The sample initializes the **SPIM** and **SPIS** instances from the board's :file:`board-config.h`.
```
eivindj-nordic:
```suggestion
The sample initializes the **SPIM** and **SPIS** instances with the pins configured in the board's :file:`board-config.h`.
```

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+LED 2
+   Toggles each time the SPI slave completes a reception.
+
+Button 2
+   Sends a message from the local SPIM to the connected SPIS.
+
+Building and running
+********************
+
+This sample is in :file:`samples/peripherals/spi/` in the |BMshort| folder structure.
```
eivindj-nordic: Let's align with other samples.
```suggestion
This sample can be found under :file:`samples/peripherals/spi/` in the |BMshort| folder structure.
```

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+
+For general instructions, see :ref:`getting_started_with_the_samples`.
```
eivindj-nordic:
```suggestion
For details on how to create, configure, and program a sample, see :ref:`getting_started_with_the_samples`.
```

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+.. rubric:: Two devices, two data paths
+
+#. Flash both devices with the same image.
+#. Wire both sets of four SPIM→SPIS signals (8 wires) **and GND** (see :ref:`bm_spi_wiring`).
+#. Open a console for each device.
+#. Press **Button 2** on either device to send to the other.
+   The sending device logs "sent", the receiving device logs "received",
+   and **LED 2** toggles on the receiver.
+
+A successful test shows the :kconfig:option:`CONFIG_SAMPLE_SPI_MSG` string in both the master completion and slave reception log entries.
```
eivindj-nordic: This is a sample, not a test. 
`Observe ...`

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,292 @@
...
+This sample is in :file:`samples/peripherals/spi/` in the |BMshort| folder structure.
+
+For general instructions, see :ref:`getting_started_with_the_samples`.
+
+Testing
+=======
+
+After building and flashing, verify the wiring configuration you chose:
+
+.. rubric:: Single device (loopback)
```
eivindj-nordic: We have tabs for this above, let's do it the same way here.

eivindj-nordic (commented):
Some overall comments not diving in to the code.

boards/nordic/bm_nrf54l15dk/include/board-config.h
PizzaAllTheWay: fixed

boards/nordic/bm_nrf54l15dk/include/board-config.h
PizzaAllTheWay: You are right, this does however mean we need a bit of refactoring of the code, because not all SoC support master and slave SPI peripherals at the same time without modifying the board (eks: nRF54LV10)

PizzaAllTheWay: Ok I added new boards and refactored code and documentation to fit with the new sample code

samples/peripherals/spi/CMakeLists.txt
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: oki, removed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,88 @@
...
+
+.. include:: /includes/spi_board_connections.txt
+
+Connect the four SPIM pins to the SPIS pins on the same board, matching each signal by name (SCK→SCK, MOSI→MOSI, MISO→MISO, CSN→CSN).
+
+.. note:: Board-specific behavior
+
+   * **nRF54L15 DK** — SPI Master pins overlap with **LED 3** (P1.14, flickers during transfers) and **Button 0** (P1.13, do not press during transfers).
+   * **nRF54LV10 DK** — The SPI Slave pins (P0.00-P0.03) are shared with one of the debugger's virtual serial ports.
+     Before running the sample, open the `Board Configurator`_ app in `nRF Connect for Desktop`_ and disable the **Connect port VCOM** entry that is mapped to pins **P0.00-P0.03**, to release these pins from the debugger.
```
eivindj-nordic: Can we use other pins for this? I'd rather not touch the board controller.

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,88 @@
...
+     Leave the other **Connect port VCOM** entry (mapped to pins **P1.04-P1.07**) enabled, since it is the virtual serial port used to read the sample's log output.
```
eivindj-nordic: Leave for testing steps.

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Testing
+=======
+
+After building and flashing the sample, test it as follows:
+
+1. Connect the four SPIM→SPIS signal pairs on the same board (see :ref:`Wiring <bm_spi_wiring>`).
+#. Open the console log and verify that ``SPI sample initialized`` appears in the log.
+#. Press **Button 2**.
+#. Observe that the log shows a master "sent" entry and a slave "received" entry, both containing the :kconfig:option:`CONFIG_SAMPLE_SPI_MSG` string.
```
eivindj-nordic:
```suggestion
#. Observe that the log shows a master "sent" entry and a slave "received" entry, both containing the string set by the :kconfig:option:`CONFIG_SAMPLE_SPI_MSG` Kconfig option.
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,153 @@
...
+
+LOG_MODULE_REGISTER(sample, CONFIG_SAMPLE_SPI_LOG_LEVEL);
+
+static nrfx_spim_t spim_inst = NRFX_SPIM_INSTANCE(BOARD_APP_SPIM_INST);
+static nrfx_spis_t spis_inst = NRFX_SPIS_INSTANCE(BOARD_APP_SPIS_INST);
+
+static uint8_t tx_buf[] = CONFIG_SAMPLE_SPI_MSG;
+static uint8_t rx_buf[sizeof(CONFIG_SAMPLE_SPI_MSG)];
+
+static void spim_handler(nrfx_spim_event_t const *p_event, void *p_context)
```
eivindj-nordic: Remove hungarian notation, here and others. 
```suggestion
static void spim_handler(nrfx_spim_event_t const *event, void *context)
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,153 @@
...
+}
+
+int main(void)
+{
+	int err;
+	nrfx_spim_config_t spim_config = NRFX_SPIM_DEFAULT_CONFIG(BOARD_APP_SPIM_PIN_SCK,
+								  BOARD_APP_SPIM_PIN_MOSI,
+								  BOARD_APP_SPIM_PIN_MISO,
+								  BOARD_APP_SPIM_PIN_CSN);
+	/* Default is 4 MHz. Lower the frequency for reliable two-device operation. */
```
eivindj-nordic: Should not be needed.

samples/peripherals/spi/README.rst
PizzaAllTheWay: I agree this isnt ideal, but its quite tricky. There are no other pins available on Port 0 on the chip itself, nRF54LV10A that sits on the DK only has P0.00-P0.04 on the chip, and thats it.

The alternative would be to move the SPIS from instance 30 to instance 20 on Port 1, which is the other SPIS peripheral instance space available. However, instance 20 on Port 1 is already used by the logger to print output for the user, and we definitely can't touch that. otherwise, the user wouldn't see any output in the logger terminal when running the sample :/

So unfortunately I dont see any less bad solution here. We are kind of stuck with using SPIS30 on P0 if we want both slave and master SPI.

Alternatively master and slave SPI could share the same instance, however that would mean the user requires 2 DKs to test the sample, I don't know if that's something we want either.

Let me know what you think about this situation. If you know how we should proceed with this.

samples/peripherals/spi/README.rst
PizzaAllTheWay: I moved the connection and wiring to testing chapter. Please check if I did it correctly

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed for `spim_handler()` and `spis_handler()`

samples/peripherals/spi/src/main.c
PizzaAllTheWay: Agree, removed the comment.

samples/peripherals/spi/README.rst
eivindj-nordic: Alright, lets keep it as is then.

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,152 @@
...
+}
+
+int main(void)
+{
+	int err;
+	nrfx_spim_config_t spim_config = NRFX_SPIM_DEFAULT_CONFIG(BOARD_APP_SPIM_PIN_SCK,
+								  BOARD_APP_SPIM_PIN_MOSI,
+								  BOARD_APP_SPIM_PIN_MISO,
+								  BOARD_APP_SPIM_PIN_CSN);
+	spim_config.frequency = NRFX_MHZ_TO_HZ(4);
```
eivindj-nordic: Why does the frequency deviate from the default? Should at least be commented, but does it need to?

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,152 @@
...
+	nrf_gpio_pin_write(BOARD_PIN_LED_2, !BOARD_LED_ACTIVE_STATE);
+
+	LOG_INF("LEDs initialized");
+
+	/* User info on startup */
+	LOG_INF("Press Button 2 to send \"%s\" over SPI", CONFIG_SAMPLE_SPI_MSG);
+	LOG_INF("Waiting for SPI master to send data...");
+
+	nrf_gpio_pin_write(BOARD_PIN_LED_0, BOARD_LED_ACTIVE_STATE);
+	LOG_INF("SPI sample initialized");
```
eivindj-nordic:
```suggestion
	nrf_gpio_pin_write(BOARD_PIN_LED_0, BOARD_LED_ACTIVE_STATE);
	LOG_INF("SPI sample initialized");
	
	/* User info on startup */
	LOG_INF("Press Button 2 to send \"%s\" over SPI", CONFIG_SAMPLE_SPI_MSG);
	LOG_INF("Waiting for SPI master to send data...");
```

eivindj-nordic (approved):
Looks good! Couple of nits :)

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: It's residue from iteration with 2 separate devices and SPI bus, where I lowered the frequency for a more robust connection. When the sample became single device loopback only I changed frequency to default. I had a comment explaining this line to showcase how to change frequency on the bus and how it correlates to noise and bandwidth, but ended up removing it.

This line is not necessary since it matches `NRFX_SPIM_DEFAULT_CONFIG` default frequency.

I removed the line.

eivindj-nordic:

nordicjm (approved):

doc/nrf-bm/links.txt
```diff
@@ -90,6 +90,7 @@
 .. _`How to flash an application`: https://docs.nordicsemi.com/bundle/nrf-connect-vscode/page/get_started/quick_debug.html#how-to-flash-an-application
 .. _`How to connect to the terminal`: https://docs.nordicsemi.com/bundle/nrf-connect-vscode/page/get_started/quick_debug.html#how-to-connect-to-the-terminal
 .. _`nRF Connect for Desktop`: https://www.nordicsemi.com/Software-and-Tools/Development-Tools/nRF-Connect-for-desktop
+.. _`Board Configurator`: https://docs.nordicsemi.com/bundle/nrf-connect-board-configurator/page/index.html
```
anhmolt: This link is broken.

samples/peripherals/spi/prj.conf
```diff
@@ -0,0 +1,20 @@
...
+
+# Clock control
+CONFIG_CLOCK_CONTROL=y
+
+# LED libraries
+CONFIG_BM_GPIOTE=y
+
+# Button libraries
+CONFIG_BM_BUTTONS=y
+CONFIG_BM_TIMER=y
```
anhmolt:
```suggestion
# Button and timer
CONFIG_BM_BUTTONS=y
CONFIG_BM_GPIOTE=y
CONFIG_BM_TIMER=y
```
The `bm_gpiote` library is for the buttons, not the LED pins.

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+	BM_IRQ_DIRECT_CONNECT(NRFX_IRQ_NUMBER_GET(BOARD_APP_SPIM_INST), IRQ_PRIO_LOWEST,
+			      spim_isr_handler, 0);
+	irq_enable(NRFX_IRQ_NUMBER_GET(BOARD_APP_SPIM_INST));
+	BM_IRQ_DIRECT_CONNECT(NRFX_IRQ_NUMBER_GET(BOARD_APP_SPIS_INST), IRQ_PRIO_LOWEST,
+			      spis_isr_handler, 0);
+	irq_enable(NRFX_IRQ_NUMBER_GET(BOARD_APP_SPIS_INST));
+
+	err = nrfx_spim_init(&spim_inst, &spim_config, spim_handler, NULL);
+	if (err) {
+		LOG_ERR("SPIM init failed: %d", err);
```
anhmolt:
```suggestion
		LOG_ERR("SPIM init failed, err %d", err);
```
Here and others.

doc/nrf-bm/links.txt
PizzaAllTheWay: Good catch, fixed

samples/peripherals/spi/prj.conf
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+
+	err = bm_buttons_enable();
+	if (err) {
+		LOG_ERR("Buttons enable failed, err %d", err);
+		goto idle;
+	}
+
+	LOG_INF("Buttons initialized");
+
+	/* LED setup */
```
anhmolt:
```suggestion
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+	nrf_gpio_cfg_output(BOARD_PIN_LED_0);
+	nrf_gpio_cfg_output(BOARD_PIN_LED_2);
+	nrf_gpio_pin_write(BOARD_PIN_LED_0, !BOARD_LED_ACTIVE_STATE);
+	nrf_gpio_pin_write(BOARD_PIN_LED_2, !BOARD_LED_ACTIVE_STATE);
+
+	LOG_INF("LEDs initialized");
```
anhmolt:
```suggestion
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+
+	nrf_gpio_pin_write(BOARD_PIN_LED_0, BOARD_LED_ACTIVE_STATE);
+	LOG_INF("SPI sample initialized");
+
+	/* User info on startup */
+	LOG_INF("Press Button 2 to send \"%s\" over SPI", CONFIG_SAMPLE_SPI_MSG);
```
anhmolt:
```suggestion
	LOG_INF("Press Button 2 to send \"%s\" over SPI", tx_buf);
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	err = nrfx_spis_buffers_set(&spis_inst, NULL, 0, rx_buf, sizeof(rx_buf));
+	if (err) {
+		LOG_ERR("SPIS buffers set failed, err %d", err);
+		goto idle;
+	}
+	LOG_INF("SPI initialized");
+
+	/* Button setup */
```
anhmolt:
```suggestion
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+		goto idle;
+	}
+
+	LOG_INF("Buttons initialized");
+
+	/* LED setup */
+	nrf_gpio_cfg_output(BOARD_PIN_LED_0);
+	nrf_gpio_cfg_output(BOARD_PIN_LED_2);
+	nrf_gpio_pin_write(BOARD_PIN_LED_0, !BOARD_LED_ACTIVE_STATE);
+	nrf_gpio_pin_write(BOARD_PIN_LED_2, !BOARD_LED_ACTIVE_STATE);
```
anhmolt: Lets move this right below `LOG_INF("SPI sample started");`

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+		goto idle;
+	}
+
+	err = bm_buttons_enable();
+	if (err) {
+		LOG_ERR("Buttons enable failed, err %d", err);
+		goto idle;
+	}
+
+	LOG_INF("Buttons initialized");
```
anhmolt:
```suggestion
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+	}
+	LOG_INF("SPI initialized");
+
+	/* Button setup */
+	err = bm_buttons_init(btn_configs, ARRAY_SIZE(btn_configs),
+			      BM_BUTTONS_DETECTION_DELAY_MIN_US);
+	if (err) {
+		LOG_ERR("Buttons init failed, err %d", err);
+		goto idle;
+	}
```
anhmolt: Lets move this right below `LOG_INF("SPI sample started");` and the LED setup.

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+static nrfx_spim_t spim_inst = NRFX_SPIM_INSTANCE(BOARD_APP_SPIM_INST);
+static nrfx_spis_t spis_inst = NRFX_SPIS_INSTANCE(BOARD_APP_SPIS_INST);
+
+static uint8_t tx_buf[] = CONFIG_SAMPLE_SPI_MSG;
+static uint8_t rx_buf[sizeof(CONFIG_SAMPLE_SPI_MSG)];
+
+static void spim_handler(nrfx_spim_event_t const *event, void *context)
+{
+	if (event->type == NRFX_SPIM_EVENT_DONE) {
+		LOG_INF("Message sent: \"%s\"", CONFIG_SAMPLE_SPI_MSG);
```
anhmolt:
```suggestion
		LOG_INF("Message sent: \"%s\"", tx_buf);
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+#include <nrfx_spim.h>
+#include <nrfx_spis.h>
+
+LOG_MODULE_REGISTER(sample, CONFIG_SAMPLE_SPI_LOG_LEVEL);
+
+static nrfx_spim_t spim_inst = NRFX_SPIM_INSTANCE(BOARD_APP_SPIM_INST);
+static nrfx_spis_t spis_inst = NRFX_SPIS_INSTANCE(BOARD_APP_SPIS_INST);
+
+static uint8_t tx_buf[] = CONFIG_SAMPLE_SPI_MSG;
+static uint8_t rx_buf[sizeof(CONFIG_SAMPLE_SPI_MSG)];
```
anhmolt: In my mind, the size of the receive buffer should not depend on the size of the message to send. Yes, it is a simple sample and it would be correct if limiting this to only do loopback, but lets do it a bit more generic.

What about adding a Kconfig for the size of this buffer. Set it to default 32 or 64 bytes?

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+		LOG_ERR("SPIS init failed, err %d", err);
+		goto idle;
+	}
+
+	err = nrfx_spis_buffers_set(&spis_inst, NULL, 0, rx_buf, sizeof(rx_buf));
+	if (err) {
+		LOG_ERR("SPIS buffers set failed, err %d", err);
+		goto idle;
+	}
+	LOG_INF("SPI initialized");
```
anhmolt:
```suggestion
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+			.pin_number = BOARD_PIN_BTN_2,
+			.active_state = BM_BUTTONS_ACTIVE_LOW,
+			.pull_config = BM_BUTTONS_PIN_PULLUP,
+			.handler = button_handler,
+		},
+	};
+
+	LOG_INF("SPI sample started");
+
+	/* SPI setup */
```
anhmolt:
```suggestion
```

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,151 @@
...
+	irq_enable(NRFX_IRQ_NUMBER_GET(BOARD_APP_SPIM_INST));
+	BM_IRQ_DIRECT_CONNECT(NRFX_IRQ_NUMBER_GET(BOARD_APP_SPIS_INST), IRQ_PRIO_LOWEST,
+			      spis_isr_handler, 0);
+	irq_enable(NRFX_IRQ_NUMBER_GET(BOARD_APP_SPIS_INST));
+
+	err = nrfx_spim_init(&spim_inst, &spim_config, spim_handler, NULL);
+	if (err) {
+		LOG_ERR("SPIM init failed, err %d", err);
+		goto idle;
+	}
```
anhmolt, line 110:
```suggestion
	err = nrfx_spim_init(&spim_inst, &spim_config, spim_handler, NULL);
	if (err) {
		LOG_ERR("SPIM init failed, err %d", err);
		goto idle;
	}
	
	/* Increase the delay between chip select being pulled low and the first clock pulse.
	 *
	 * This allows an SPI target device with a long wakeup time a chance to get ready before
	 * data is transferred. This is necessary when using the SPIS peripheral on devices
	 * such as the nRF54L, where the high-frequency clock is powered down during sleep and needs
	 * time to start before the SPIS peripheral can function correctly. An alternative is
	 * keeping the high-frequency clock running at all times.
	 */
	spim_inst.p_reg->IFTIMING.CSNDUR = 80;
	
```

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,83 @@
...
+   The pins are defined in the board's :file:`board-config.h` header (``BOARD_APP_SPIM_*`` and ``BOARD_APP_SPIS_*``).
+
+   .. include:: /includes/spi_board_connections.txt
+
+   .. note:: Board-specific behavior
+
+      * **nRF54L15 DK** — SPI Master pins overlap with **LED 3** (P1.14, flickers during transfers) and **Button 0** (P1.13, do not press during transfers).
+      * **nRF54LV10 DK** — The SPI Slave pins (P0.00-P0.03) are shared with one of the debugger's virtual serial ports.
+        Before running the sample, open the `Board Configurator`_ app in `nRF Connect for Desktop`_ and disable the **Connect port VCOM** entry that is mapped to pins **P0.00-P0.03**, to release these pins from the debugger.
+        Leave the other **Connect port VCOM** entry (mapped to pins **P1.04-P1.07**) enabled, since it is the virtual serial port used to read the sample's log output.
```
anhmolt: Thinking we could move this part containing the wiring setup to the `Requirements` section. After tables with the supported boards. Then explain both the two-device and single-device loopback setup similar to how it has been done in `samples/peripherals/lpuarte/README.rst`.

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: ok, fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: ok, fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,83 @@
...
+********************
+
+This sample can be found under :file:`samples/peripherals/spi/` in the |BMshort| folder structure.
+
+For details on how to create, configure, and program a sample, see :ref:`getting_started_with_the_samples`.
+
+Testing
+=======
+
+After building and flashing the sample, test it as follows:
```
anhmolt:
```suggestion
After building and programming the sample, test it as follows:
```

samples/peripherals/spi/src/main.c
PizzaAllTheWay: makes sense, fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/src/main.c
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: Agree. I rewrote Testing and Requirements chapter to be better aligned with other samples that test with 2 different setup options, like LPUATE. Take a look and let me know what you think.

samples/peripherals/spi/src/main.c
PizzaAllTheWay: Makes sense I rewrote your suggested comment. Please take a look, and see if it makes sense

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,157 @@
...
+	if (event->type == NRFX_SPIM_EVENT_DONE) {
+		LOG_INF("Message sent: \"%s\"", tx_buf);
+	}
+}
+
+static void spis_handler(nrfx_spis_event_t const *event, void *context)
+{
+	if (event->evt_type == NRFX_SPIS_XFER_DONE) {
+		LOG_INF("Message received (%u bytes):", event->rx_amount);
+		LOG_HEXDUMP_INF(rx_buf, event->rx_amount, "Message");
```
anhmolt, line 37:
```suggestion
		LOG_INF("Message received (%u bytes): \"%.*s\"",
			event->rx_amount, event->rx_amount, rx_buf);
```

samples/peripherals/spi/src/main.c
bjorn-tore-taraldsen: I think we should be more specific on the problem we solve. "SPI Slave running on as a software service on a microcontroller,..... "

samples/peripherals/spi/src/main.c
```diff
@@ -0,0 +1,157 @@
...
+		LOG_ERR("Buttons enable failed, err %d", err);
+		goto idle;
+	}
+
+	nrf_gpio_pin_write(BOARD_PIN_LED_0, BOARD_LED_ACTIVE_STATE);
+	LOG_INF("SPI sample initialized");
+
+	/* User info on startup */
+	LOG_INF("Press Button 2 to send \"%s\" over SPI", tx_buf);
+	LOG_INF("Waiting for SPI master to send data...");
```
anhmolt, line 147:
```suggestion
	LOG_INF("Waiting for SPI controller to send data...");
```
Lets use "SPI controller" instead of "SPI master" and "SPI target" instead of "SPI slave". That should harmonize with the product spec.'s titles for SPIM and SPIS.

anhmolt (commented):
Here are some more suggestions: https://github.com/PizzaAllTheWay/sdk-nrf-bm/compare/spi-sample...anhmolt:sdk-nrf-bm:anhmolt-spi-sample-suggestions

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,102 @@
...
+Testing
+=======
+
+You can test this sample by performing the following steps:
+
+#. Compile and program the application.
+#. Open the console log and verify that ``SPI sample initialized`` appears in the log.
+#. Press **Button 2**.
+#. Observe that the log shows a master "sent" entry and a slave "received" entry, both containing the string set by the :kconfig:option:`CONFIG_SAMPLE_SPI_MSG` Kconfig option.
+#. Observe that **LED 2** toggles on each completed reception.
```
anhmolt, line 102: This could use some updates to facilitate for the two-device test-case.

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -248,6 +249,11 @@ NFC samples
 
 No changes since the latest nRF Connect SDK Bare Metal release.
 
+Peripheral samples
```
b-gent: This entry duplicates the 'Peripheral samples' above. pls remove

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+   .. group-tab:: MCUboot board variants
+
+      The following board variants have DFU capabilities:
+
+      .. include:: /includes/supported_boards_all_mcuboot_variants_s115.txt
+
+      .. include:: /includes/supported_boards_all_mcuboot_variants_s145.txt
+
+The sample requires the following pin wiring:
```
b-gent, line 45:
```suggestion
      .. include:: /includes/supported_boards_all_mcuboot_variants_s145.txt

Wiring
*******

The sample requires the following pin wiring:
```

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,96 @@
...
+.. include:: /includes/spi_board_connections.txt
+
+.. note:: Board-specific behavior
+
+   * **nRF54L15 DK** — The SPI controller pins overlap with **LED 3** (P1.14, flickers during transfers) and **Button 0** (P1.13, do not press during transfers).
+   * **nRF54LV10 DK** — The SPI target pins (P0.00-P0.03) are shared with one of the debugger's virtual serial ports.
+     Before running the sample, open the `Board Configurator`_ app in `nRF Connect for Desktop`_ and disable the **Connect port VCOM** entry that is mapped to pins **P0.00-P0.03**, to release these pins from the debugger.
+     Leave the other **Connect port VCOM** entry (mapped to pins **P1.04-P1.07**) enabled, since it is the virtual serial port used to read the sample's log output.
+
+Overview
```
b-gent: Move Overview before Wiring.

(as per sample template https://github.com/nrfconnect/sdk-nrf/blob/main/doc/nrf/templates/sample_README.rst#wiring)

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,96 @@
...
+The sample requires the following pin wiring:
+
+Single-device loopback setup:
+   Wire the four SPIM pins to the SPIS pins on the same board, matching each signal by name (SCK→SCK, MOSI→MOSI, MISO→MISO, CSN→CSN).
+
+Two-device setup:
+   Wire the four SPIM pins on device 1 to the SPIS pins on device 2, matching each signal by name (SCK→SCK, MOSI→MOSI, MISO→MISO, CSN→CSN).
+   Optionally, wire the four SPIS pins on device 1 to the SPIM pins on device 2 to test SPIM and SPIS on both devices.
+
+   .. note::
```
b-gent: If I undrestand correctly, this note applies to both wiring configurations, so it should not be indented inside the second point.

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+LED 0:
+   Lit when the device is initialized.
+
+LED 2:
+   Toggles when the SPIS peripheral completes a reception.
+
+Button 2:
+   Send string from the local SPIM to a connected SPI target.
+
```
b-gent, line 80: Optionally, a 'Configuration' section can be added here with an explanation of the Kconfig options mentioned in this readme.

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,96 @@
...
+Building and running
+********************
+
+This sample can be found under :file:`samples/peripherals/spi/` in the |BMshort| folder structure.
+
+For details on how to create, configure, and program a sample, see :ref:`getting_started_with_the_samples`.
+
+Testing
```
b-gent, line 101: 1. We need to mention to complete the steps in 'Wiring' first, depending on the selected config.
2. Is the Testing exactly the same for each wiring config? We could have tabs for the two different options.

Suggestion for review (with some AI assistance):

```suggestion
Testing
=======

.. tabs::

   .. group-tab:: Single-device loopback

      Test the sample on one development kit, with its SPI controller wired to its own SPI target.

      1. Wire the SPIM pins to the SPIS pins on the same kit, as described in `Wiring`_.
      #. Compile and program the application.
      #. Connect to the kit with a terminal emulator (for example, the `Serial Terminal app`_).
      #. Observe that the ``SPI sample initialized`` message is printed.
      #. Press **Button 2**.
      #. Observe that the terminal prints ``Message sent`` followed by the string set by the :kconfig:option:`CONFIG_SAMPLE_SPI_MSG` Kconfig option (``Hello World!`` by default).
      #. Observe that the terminal prints ``Message received``, the number of received bytes, and the same string.
      #. Observe that **LED 2** toggles.

   .. group-tab:: Two-device setup

      Test the sample on two development kits, with the SPI controller on device 1 wired to the SPI target on device 2.

      1. Wire device 1 and device 2 as described in `Wiring`_.
      #. Compile and program the application on both kits.
      #. Connect to both kits with a terminal emulator (for example, the `Serial Terminal app`_).
         In the following steps, these are referred to as the **device 1 terminal** and the **device 2 terminal**.
      #. Observe that both terminals print the ``SPI sample initialized`` message.
      #. Press **Button 2** on device 1.
      #. Observe that the device 1 terminal prints ``Message sent`` followed by the string set by the :kconfig:option:`CONFIG_SAMPLE_SPI_MSG` Kconfig option (``Hello World!`` by default).
      #. Observe that the device 2 terminal prints ``Message received``, the number of received bytes, and the same string.
      #. Observe that **LED 2** toggles on device 2.
      #. If you also wired the second data path, press **Button 2** on device 2 and repeat the previous three checks with the roles of the two devices reversed.
```

b-gent (changes requested):

doc/nrf-bm/release_notes/release_notes_changelog.rst
PizzaAllTheWay: good catch fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: No, the indentation is correct, the note only applies to two-device, not single-device loopback setup.

In loopback, SPIM and SPIS are on the same chip, so they share one clock. The moment SPIM turns that clock on to send data, it's already on for SPIS as well, both sides wake up together because they're really the same device. There's no waiting needed since nothing was asleep to begin with.

In the two-device setup, each board manages its own power and clock separately. The receiving board can be asleep with its clock off, and the sending board has no way of knowing this, it can't see inside the other board. If data is sent right away, the receiving board might still be waking up and miss the start of the message.

That's what the delay is for. After the sender raises chip select, it waits briefly before sending real data, giving the other board time to fully wake up so it doesn't miss anything. This only matters in the two-device case, since that's the only setup where one side can be asleep while the other is talking.

samples/peripherals/spi/README.rst
PizzaAllTheWay: Agree, added 'Configuration' section now, please take a look

samples/peripherals/spi/README.rst
PizzaAllTheWay: Yes, the testing steps are the same for both wiring setups. That said, separating them into tabs makes more sense and would align with other samples like `BLE_NUS`. Your suggested change looks good, I applied it with some minor modifications, please take a look

samples/peripherals/spi/README.rst
eivindj-nordic: I think @b-gent has a point. Yes the delay is only required in a two device setup where the SPIS requires extra wakeup time, though we do not know that at compile time. Hence the extra delay is present also in one device setup. So I think it makes sense to reduce the indentation level.  I'm also not sure whether we need to mention the alternative of keeping the high-frequency clock always on on the nRF54l when acting as a SPI device.

PizzaAllTheWay: Ok, I removed the indentation and changed the text

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Overview
+********
+
+The sample initializes the **SPIM** and **SPIS** instances with the pins configured in the :file:`board-config.h` header.
+When **Button 2** is pressed, a configurable string (:kconfig:option:`CONFIG_SAMPLE_SPI_MSG`) is sent from the local SPI controller to a connected SPI target.
+The local SPI target receives data from a connected SPI controller, logs the received data and toggles **LED 2**.
+
+Wiring
+*******
```
anhmolt:
```suggestion
Wiring
******
```

samples/peripherals/spi/README.rst
```diff
@@ -0,0 +1,134 @@
...
+********************
+
+This sample can be found under :file:`samples/peripherals/spi/` in the |BMshort| folder structure.
+
+For details on how to create, configure, and program a sample, see :ref:`getting_started_with_the_samples`.
+
+Testing
+=======
+
+You can test the sample in two ways, depending on the selected wiring: the single-device loopback setup or the two-device setup.
```
anhmolt, line 104:
```suggestion
You can test the sample in two ways depending on the selected wiring:
```

The tab titles are visible below, so I don't think we need to repeat here.

samples/peripherals/spi/README.rst
anhmolt: Looks good!

anhmolt (approved):

samples/peripherals/spi/README.rst
PizzaAllTheWay: Good catch, fixed

samples/peripherals/spi/README.rst
PizzaAllTheWay: I am not sure if we should remove this. I tried to align README with how it is done other places where 2 testing ways are possible (for example `ble_nus`). There we state both options:
https://github.com/nrfconnect/sdk-nrf-bm/blob/main/samples/bluetooth/ble_nus/README.rst?plain=1#L75

b-gent (approved):

bjorn-tore-taraldsen (approved):

PR #763, opened 2026-04-21
tests: lib: bm_timer: add ztest suite

lemrey:
Add `ztest` suite for `bm_timer` running on hardware.

### DIFF
--- a/tests/lib/bm_timer/prj.conf
+++ b/tests/lib/bm_timer/prj.conf
@@ -1,0 +1,14 @@
+#
+# Copyright (c) 2026 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Enable the ZTEST framework
+CONFIG_ZTEST=y
+
+# Module under test
+CONFIG_BM_TIMER=y
+
+# Logging for test output
+CONFIG_LOG=y
--- a/tests/lib/bm_timer/src/main.c
+++ b/tests/lib/bm_timer/src/main.c
@@ -1,0 +1,213 @@
+/*
+ * Copyright (c) 2026 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <errno.h>
+#include <stdint.h>
+#include <string.h>
+#include <zephyr/kernel.h>
+#include <zephyr/sys/atomic.h>
+#include <zephyr/ztest.h>
+
+#include <bm/bm_timer.h>
+
+/* Timer duration that is comfortably above BM_TIMER_MIN_TIMEOUT_TICKS but short enough
+ * to keep the tests fast. ~50 ms is plenty for both single shot and repeated tests.
+ */
+#define TEST_TIMER_MS         50U
+#define TEST_TIMER_TICKS      BM_TIMER_MS_TO_TICKS(TEST_TIMER_MS)
+
+/* Slack to allow the timer to fire before the test thread checks the result. */
+#define TEST_TIMER_WAIT_MS    (TEST_TIMER_MS + 50U)
+
+/* Magic value used to verify the context pointer is forwarded to the handler. */
+#define TEST_CONTEXT_MAGIC    ((void *)0xDEADBEEFUL)
+
+/* Shared test state updated from the timer expiry handler (interrupt context). */
+static atomic_t handler_call_count;
+static void *last_context;
+
+static void timeout_handler(void *context)
+{
+	last_context = context;
+	atomic_inc(&handler_call_count);
+}
+
+static void reset_handler_state(void)
+{
+	atomic_set(&handler_call_count, 0);
+	last_context = NULL;
+}
+
+static void before(void *fixture)
+{
+	ARG_UNUSED(fixture);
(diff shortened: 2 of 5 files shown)


tests/lib/bm_timer/src/main.c
```diff
@@ -0,0 +1,245 @@
...
+	zassert_ok(err, "bm_timer_init failed, err %d", err);
+	zassert_equal(timer.mode, BM_TIMER_MODE_SINGLE_SHOT, "unexpected mode");
+	zassert_equal_ptr(timer.handler, timeout_handler, "unexpected handler");
+
+	memset(&timer, 0, sizeof(timer));
+
+	err = bm_timer_init(&timer, BM_TIMER_MODE_REPEATED, timeout_handler);
+	zassert_ok(err, "bm_timer_init failed, err %d", err);
+	zassert_equal(timer.mode, BM_TIMER_MODE_REPEATED, "unexpected mode");
+	zassert_equal_ptr(timer.handler, timeout_handler, "unexpected handler");
```
eivindj-nordic: Split in two?

tests/lib/bm_timer/src/main.c
```diff
@@ -0,0 +1,245 @@
...
+	zassert_ok(err, "bm_timer_init failed, err %d", err);
+
+	err = bm_timer_start(&timer, 0, NULL);
+	zassert_equal(err, -EINVAL, "expected -EINVAL for 0 ticks, got %d", err);
+
+	if (BM_TIMER_MIN_TIMEOUT_TICKS > 1U) {
+		err = bm_timer_start(&timer, BM_TIMER_MIN_TIMEOUT_TICKS - 1U, NULL);
+		zassert_equal(err, -EINVAL,
+			      "expected -EINVAL for ticks below min, got %d", err);
+	}
```
eivindj-nordic, line 80: Separate test?

tests/lib/bm_timer/src/main.c
```diff
@@ -0,0 +1,245 @@
...
+			  "context not propagated to handler");
+}
+
+ZTEST(bm_timer, test_bm_timer_restart)
+{
+	int err;
+	struct bm_timer timer;
+	atomic_val_t count;
+
+	err = bm_timer_init(&timer, BM_TIMER_MODE_SINGLE_SHOT, timeout_handler);
```
eivindj-nordic, line 193: Should this be repeated? With single shot it only triggers once, so this does not verify if the second start does nothing.

lemrey: We verify that it fires only once, so it is restarted.

tests/lib/bm_timer/src/main.c
lemrey: Updated

tests/lib/bm_timer/src/main.c
lemrey: Actually, this is tested properly below already. I removed this test.

tests/lib/bm_timer/src/main.c
eivindj-nordic: My point is that it could just as well be the first timer_start that triggers once, then the next call to start is ignored. The test would not separate the two, i.e. we are not verifying that the timer is in fact restarted.

lemrey: Ah, I see. I updated the test to include a context when we restart the timer, so we can check against that.

tests/lib/bm_timer/prj.conf
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright (c) 2026 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Enable the ZTEST framework
+CONFIG_ZTEST=y
+CONFIG_SOFTDEVICE=y
```
anhmolt: Is `CONFIG_SOFTDEVICE=y` needed to enable the ZTEST framework?

anhmolt (approved):

tests/lib/bm_timer/prj.conf
lemrey: no,but it must be enabled in every project (it is not enabled by default, and no board enables it via defconfig)

anhmolt: Right, then it appear strange that it seems to be placed under `# Enable the ZTEST framework`. Could consider a newline.

lemrey: or.. the test must be built without sysbuild; that's where the warning comes from. perhaps it's okay to remove this then provided that the test is compiled with `--no-sysbuild`.

eivindj-nordic: I think as long as we only have SoftDevice board variants we should enable the Kconfig.

eivindj-nordic (approved):

tests/lib/bm_timer/prj.conf
anhmolt: Test passes for me locally with disabled `CONFIG_SOFTDEVICE`, both with and without `--no-sysbuild`. But I see a warning when building with sysbuild. When running with `west twister`, the test passes. So I'm fine with any.

eivindj-nordic:

nordicjm (approved):

PR #769, opened 2026-04-22
mgmt: mcumgr: grp: img: mgmt: bm_img_mgmt: dont write empty ring buf

eivindj-nordic:
Do not call bm_storage_write if the claim returns a length of 0.


PizzaAllTheWay (approved):
LGTM

anhmolt (approved):

lemrey (commented):
When could it happen that we retrieve zero bytes from the buffer? Or is it the rounding that causes that?

eivindj-nordic:
> When could it happen that we retrieve zero bytes from the buffer? Or is it the rounding that causes that?

At the end when the entire image is received, you will claim 0 bytes at the last write event.

lemrey (approved):
This is needed only when `last_data` is `true`, otherwise rounding down can't round to zero given that we always claim at least `storage_wear_unit` bytes.

michalek-no (approved):

PR #771, opened 2026-04-22
license: Add softdevice license

nicu1989:
Define license for the .hex artifacts inside softdevice.

### DIFF
--- a/components/softdevice/license.txt
+++ b/components/softdevice/license.txt
@@ -1,0 +1,7 @@
+/*
+ * Copyright (c) 2026 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ *
+ * NCS-SBOM-Apply-To-File: ./**/*_softdevice.hex
+ */


components/softdevice/license.txt
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright (c) 2026 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ *
+ * NCS-SBOM-Apply-To-File: ./**/*_softdevice.hex
+ */
```
hermabe, line 7: - We already have a license file (for example https://github.com/nrfconnect/sdk-nrf-bm/blob/ae28e495386f4c2112f317e5a32e71322e60c20b/components/softdevice/nrf54l/s115/s115_10.0.0-1.prototype_license-agreement.txt). Is this one needed?
- The license applies to all SoftDevice files, not just the hexfiles but the headers and docs as well

bjorn-tore-taraldsen: I think this file is added to make the software license available for tools like SBOM (i.e. machine readable format).
But this license file should be part of the SD deliverables from Dragoon and not a file added by us (NCS BM).

nicu1989: Yes, that`s the license that applies to at least the .hex files as for my understanding as well. Source code(headers, c files, scripts, etc) should have the license defined inside the files.

The reason for creating the license.txt file is to allow the [SBOM tool](https://github.com/nrfconnect/sdk-nrf/blob/main/scripts/west_commands/sbom/README.rst#overview) to apply the correct license to those .hex file when generating the report. This is similar to how its done in the case of .a libs: https://github.com/nrfconnect/sdk-nrfxlib/blob/main/nfc/license.txt

eivindj-nordic: @hermabe Is this something you can add in your deliverables? 
I think that is cleaner than adding it as a separate file here.

hermabe: Yes, we could do that. Is this required for the 2.0 release?

Created DRGN-28347 for tracking

nicu1989: @eivindj-nordic @hermabe  not really sure how that it will work. For example if I build sdk-nrf-bm/tree/main/samples/bluetooth/hello_softdevice with sysbuild I would end up with 2 domains, the app domain and the softdevice domain. I would then point the SBOM tool to analyze the entire build dir. And with the help of tags `NCS-SBOM-Apply-To-File` and `SPDX-License-Identifier` the license is associated with the binary.

@carlescufi what do you think?

hermabe: There may be some more changes needed to the ncs sbom tool. Running it on a sample built on this PR or main fails since there is no elf/map file for the SoftDevice:
```
heb1:~/repos/ncs/nrf-bm$ west ncs-sbom -d build/softdevice
FATAL ERROR: Cannot find map file for "zephyr/zephyr.elf" in build directory "build/softdevice". Expected location "build/softdevice/zephyr/zephyr.map".
```

eivindj-nordic: @bjorn-tore-taraldsen to comment on the timeline. 
@nicu1989 The SoftDevice deliverable includes both the SoftDevice hex files in addition to the documentation etc. for the SoftDevice, basically all files in the components/softdevice folder. So in practice it will not be much different than what is added in this PR, other than it will be part of the SoftDevice deliverable and it might be placed some layers below, for example have one license.txt file per SoftDevice variant (up to @hermabe and his team to decide what is best).

bjorn-tore-taraldsen: @hermabe  I would like to get the license files in place for this release.

nicu1989: > There may be some more changes needed to the ncs sbom tool. Running it on a sample built on this PR or main fails since there is no elf/map file for the SoftDevice:
> 
> ```
> heb1:~/repos/ncs/nrf-bm$ west ncs-sbom -d build/softdevice
> FATAL ERROR: Cannot find map file for "zephyr/zephyr.elf" in build directory "build/softdevice". Expected location "build/softdevice/zephyr/zephyr.map".
> ```

forgot to mention the sbom update PR: https://github.com/nrfconnect/sdk-nrf/pull/28328

hermabe: Adding it to the existing license file worked for me. Unless there a reason to not do this I will add it to the next softdevice push (Example for s115 for 54L https://github.com/nrfconnect/sdk-nrf-bm/compare/main...hermabe:sdk-ncs-lite:wip/sd_sbom)

eivindj-nordic:
Included in #775

PR #773, opened 2026-04-23
firmware_loader, mcuboot_recovery_entry: improved adv name persistency over two stage updates

anhmolt:
When doing update of SoftDevice+firmware_loader followed by application upload, the firmware loader advertising name set by the "old" application would only be used by the "old" firmware loader.

After having updated SoftDevice+firmware_loader and the next step is upload of application image, the name set by the "old" application would already be cleared by the old firmware loader and the new firmware loader would always use the default name.

Improve this by moving the responsibility of clearing the advertising name from firmware loader startup to application startup. The configured advertising name would then persist over the full two stage update.

### DIFF
--- a/doc/nrf-bm/release_notes/release_notes_changelog.rst
+++ b/doc/nrf-bm/release_notes/release_notes_changelog.rst
@@ -82,6 +82,8 @@ DFU
      New solution is enabled by the :kconfig:option:`CONFIG_BM_FLAT_SETTINGS_BLUETOOTH_NAME` Kconfig option.
      It empploys newly added Inter application RAM Clipboard storage.
    * Support for nRF54L15, nRF54L10, and nRF54L05 SoCs is no longer experimental and is now fully supported and ready to be used.
+   * Firmware loader to not clear the retained DFU Device Bluetooth name after loading it from retained memory.
+     This allows the advertising name to persist after updating SoftDevice and/or firmware loader, or if the application image validation fails and the device reboots into firmware loader.
 
 Removed:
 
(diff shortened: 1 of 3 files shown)


nvlsianpu (approved):

anhmolt:
Updated the changelog.

Also added a second commit to cleanup an incorrect comment and tweak the log level for a print.

eivindj-nordic (approved):

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -418,6 +420,7 @@ DFU samples
      To enable the size-optimized configuration, set :makevar:`FILE_SUFFIX` to ``size_opt`` when building the sample.
    * The :ref:`ble_mcuboot_recovery_entry_sample` migrates to new soultion for Setting up DFU Device Bluetooth name remotely.
      See the :ref:`ug_dfu` page for details.
+   * The :ref:`ble_mcuboot_recovery_entry_sample` to clear the retained DFU Device Bluetooth name on boot.
```
b-gent, line 423:
```suggestion
   * The :ref:`ble_mcuboot_recovery_entry_sample` to clear the retained DFU Device Bluetooth name on boot.
     This was previously done by the firmware loader.
```
just a suggestion to make things clearer, unless my understanding is wrong here

b-gent (approved):

doc/nrf-bm/release_notes/release_notes_changelog.rst
anhmolt: You are on the ball. Updated with your suggestion.

PR #777, opened 2026-04-24
doc: Update of SW maturity table

bjorn-tore-taraldsen:
Update Software maturity table to reflect the current status of the features.

### DIFF
--- a/doc/nrf-bm/release_notes/bm_software_maturity.rst
+++ b/doc/nrf-bm/release_notes/bm_software_maturity.rst
@@ -89,12 +89,12 @@ A device is considered supported if at least one sample or component has been de
      - Supported from v1.0.0
    * - nRF54L05
      - Supported from v1.0.0
-   * - nRF54LV10
-     - Experimental
    * - nRF54LM20A
-     - Experimental
+     - Supported from v2.0.0
    * - nRF54LS05B
-     - Experimental
+     - Supported from v2.0.0
+   * - nRF54LV10A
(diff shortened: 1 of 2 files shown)


doc/nrf-bm/release_notes/bm_software_maturity.rst
```diff
@@ -89,12 +89,12 @@ A device is considered supported if at least one sample or component has been de
      - Supported from v1.0.0
    * - nRF54L05
      - Supported from v1.0.0
-   * - nRF54LV10
-     - Experimental
    * - nRF54LM20A
-     - Experimental
+     - Supported from v2.0.0
```
b-gent, line 93: We don't have to state "from v2.0.0" , the maturity table page is inside of the release documentation so it will only appear as Supported in the 2.0.0 docs onwards.

b-gent (changes requested):

doc/nrf-bm/release_notes/bm_software_maturity.rst
bjorn-tore-taraldsen: @b-gent:
I had the same for nRF54L series when we released v1.0. 
My thought was that we was more explicitly expressed when they was introduced, when looking back on releases.
But I'm okay removing it, but then we should remove it for all

b-gent:
This changes should be now reflected in the doc PR here -> https://github.com/nrfconnect/sdk-nrf-bm/pull/779/changes

This one can be closed

PR #778, opened 2026-04-27
bm_zms: fix a bug in case of read after a write

rghaddab:
When a read is triggered immediately after a write it can start just after a sector is closed and picks up the wrong address to start the search from.
Fix this by storing a backup read address before closing a sector.

### DIFF
--- a/subsys/fs/bm_zms/bm_zms.c
+++ b/subsys/fs/bm_zms/bm_zms.c
@@ -1140,6 +1140,12 @@ static int zms_compute_prev_addr(struct bm_zms_fs *fs, uint64_t *addr)
 	struct zms_ate empty_ate;
 	struct zms_ate close_ate;
 
+	if (SECTOR_OFFSET(*addr) >= (fs->sector_size - 2 * fs->ate_size)) {
+		LOG_ERR("Something went wrong in computing previous ate address, addr is 0x%llx",
+			*addr);
+		return -EIO;
+	}
+
 	*addr += fs->ate_size;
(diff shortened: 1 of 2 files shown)


subsys/fs/bm_zms/bm_zms.c
```diff
@@ -2242,13 +2250,21 @@ ssize_t bm_zms_read_hist(struct bm_zms_fs *fs, uint32_t id, void *data, size_t l
 	wlk_addr = fs->ate_wra;
 #endif
 
+	end_search_addr = fs->ate_wra;
+	if (SECTOR_OFFSET(wlk_addr) >= (fs->sector_size - 2 * fs->ate_size)) {
+		/* we are maybe in the middle of a GC */
+		wlk_addr = fs->ate_ra;
+		end_search_addr = wlk_addr;
+	}
+	printk("wlk_addr 0x%llx, end_search_addr 0x%llx\n", wlk_addr, end_search_addr);
```
anhmolt: This should be removed, or did you intend a `LOG_DBG(...)`? It seems this spams so much that the CI tests are not able to find the logs they are looking for.
```suggestion
```

anhmolt (commented):
This fixes the issues I have been seeing. See and address my one comment and we will try to merge this tomorrow.

We need a changelog entry, but Iet us take that in https://github.com/nrfconnect/sdk-nrf-bm/pull/779.

subsys/fs/bm_zms/bm_zms.c
rghaddab: No that shouldn't be part of the PR at all

anhmolt (approved):

lemrey (approved):

PR #780, opened 2026-04-28
bluetooth: hrs: change hrm_encode return type to uint16_t

PizzaAllTheWay:
hrm_encode() returned uint8_t, but both places that use it expect a uint16_t (ble_gatts_attr_t.init_len and ble_gatts_hvx_params_t.p_len from the SoftDevice API).

In practice the length never goes above 255, so this was not a real bug, just an inconsistency that forced the compiler to silently widen the value. Changing the return type (and the internal len variable) to uint16_t makes everything line up.

No functional change.


eivindj-nordic (approved):

eivindj-nordic:
Static function, no changelog required.

anhmolt (approved):

PR #785, opened 2026-04-30
doc: post-release doc updates

b-gent:
Typical doc updates after release.
Additionally, corrected a statement about support of NFC on nRF54LM20A.


eivindj-nordic:
Should doc/shortcuts and doc/versions.json be updated as well etc.?

b-gent:
> Should doc/shortcuts and doc/versions.json be updated as well etc.?

let's leave versions.json as it is for now , it only affects the preview server and is currently bugged, I will fix and update it later.
I updated the shortcuts.txt , though I think we don't really use that |version| shortcut anywhere

peknis (approved):

anhmolt (approved):

bjorn-tore-taraldsen (approved):

PR #786, opened 2026-04-30
tests: extend timeout to 600 seconds

fundakol:
Extended twister timeout for samples where additional boards has been added.

### DIFF
--- a/tests/subsys/bootloader/upgrade/mcuboot_recovery_retention/testcase.yaml
+++ b/tests/subsys/bootloader/upgrade/mcuboot_recovery_retention/testcase.yaml
@@ -16,7 +16,7 @@ tests:
     tags:
       - pytest
       - dfu
-    timeout: 300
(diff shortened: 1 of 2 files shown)


gchwier (approved):

jnsgsr (approved):

lemrey (commented):
Does that mean that these test take up to 10 minutes to execute each?

fundakol:
> Does that mean that these test take up to 10 minutes to execute each?

Each scenario may contain a few tests written in pytest. 600 seconds timeout is for all those tests. A lower value is usually sufficient, but sometimes, under heavy load on the test node, the timeout is insufficient and the entire scenario ends with a timeout error. 600 seconds gives us more confidence.

lemrey:
Thanks for clarifying

nvlsianpu (approved):

PR #791, opened 2026-05-08
lib: bluetooth: scan: Make name and address filter const

asilz:
Currently, the name and address filter members in the `ble_scan_filter_data` struct are not `const` even though they are never modified in the scan library. I think the API should guarantee that they are not modified since I do not see a reason for why we want to modify in the library


eivindj-nordic:
You can add release notes here: https://github.com/nrfconnect/sdk-nrf-bm/blob/main/doc/nrf-bm/release_notes/release_notes_changelog.rst

eivindj-nordic (approved):

lemrey (approved):

PizzaAllTheWay (approved):
LGTM

b-gent (approved):

PR #788, opened 2026-05-05
lib: bluetooth: enforce app PHY mask and support multi-PHY config

PizzaAllTheWay:
Fix BLE PHY handling in `ble_conn_params` so application/project PHY preferences are respected consistently, including peer-initiated PHY update requests.

The change:

- Applies an app-defined PHY allow-mask on top of SoftDevice-supported PHYs before PHY updates.
- Converts PHY Kconfig from single-choice behavior to multi-select options (for example `1M + 2M`).
- Replaces recursive `NRF_ERROR_RESOURCES` retry in PHY update with a bounded non-recursive retry path.
- Improves PHY logs by printing human-readable PHY names/bitmasks and clearer wording.
- Updates documentation to describe the new auto vs allow-mask behavior and coded PHY limitations in BM SoftDevice context (S115/S145).

PR #794, opened 2026-05-11
lib: bluetooth: ble_conn_params: fix PHY get when update in progress + support multi-PHY config

anhmolt:
* Fix PHY mode get when PHY update is in progress
* Fix `radio_phy_mode_update` function to use the preferred PHY mask when calling `sd_ble_gap_phy_update`, and not a combination of `BLE_GAP_PHYS_SUPPORTED` and the peer's preferred PHY mask. Which was incorrect.
* Return invalid param if `ble_conn_params_phy_radio_mode_set` is called with SoftDevice unsupported PHY mode
* Fix resource error fallback PHY to the current/active PHY mode.
* Added support for selecting more than one PHY mode when setting the PHY mode preference with Kconfig.

Alternative to https://github.com/nrfconnect/sdk-nrf-bm/pull/788.

### DIFF
--- a/lib/bluetooth/ble_conn_params/phy_mode.c
+++ b/lib/bluetooth/ble_conn_params/phy_mode.c
@@ -32,31 +32,35 @@ BUILD_ASSERT(CONFIG_BLE_CONN_PARAMS_PHY == BLE_GAP_PHY_AUTO ||
 static void radio_phy_mode_update(uint16_t conn_handle, int idx)
 {
 	uint32_t nrf_err;
-	ble_gap_phys_t phys = links[idx].preferred;
 	struct ble_conn_params_evt app_evt = {
 		.evt_type = BLE_CONN_PARAMS_EVT_ERROR,
 		.conn_handle = conn_handle,
 	};
 
-	nrf_err = sd_ble_gap_phy_update(conn_handle, &phys);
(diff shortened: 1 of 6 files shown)


lib/bluetooth/ble_conn_params/phy_mode.c
```diff
@@ -174,6 +174,11 @@ uint32_t ble_conn_params_phy_radio_mode_set(uint16_t conn_handle, ble_gap_phys_t
 		return NRF_ERROR_INVALID_PARAM;
 	}
 
+	if ((phy_pref.tx_phys & ~BLE_GAP_PHYS_SUPPORTED) ||
+	    (phy_pref.rx_phys & ~BLE_GAP_PHYS_SUPPORTED)) {
+		return NRF_ERROR_INVALID_PARAM;
+	}
```
eivindj-nordic, line 209: Group with previous if?

eivindj-nordic (approved):

lib/bluetooth/ble_conn_params/phy_mode.c
anhmolt: I'm thinking this is cleaner.

PizzaAllTheWay (approved):
LGTM

b-gent (approved):

PR #798, opened 2026-05-19
manifest: update sdk-nrf to latest main

eivindj-nordic:
Update sdk-nrf to e611a7a75c4380f05716c6f5a984b2e1ff9c589f.

test_sdk_dfu: PR-814


nordicjm (commented):
all fixed-partitions should be changes to mapped partitions as per zephyr 4.4 migration guide

lib/bluetooth/peer_manager/modules/peer_data_storage.c
```diff
@@ -11,6 +11,7 @@
 #include <zephyr/logging/log.h>
 #include <zephyr/sys/util.h>
 #include <zephyr/sys/atomic.h>
+#include <zephyr/storage/flash_map.h>
```
anhmolt:
```suggestion
#include <zephyr/storage/flash_map.h>
#include <zephyr/sys/util.h>
#include <zephyr/sys/atomic.h>
```

eivindj-nordic:
> all fixed-partitions should be changes to mapped partitions as per zephyr 4.4 migration guide

Updated DTS and PARTITION macros. 

We also need a patch for MCUmgr after the changes in https://github.com/zephyrproject-rtos/zephyr/pull/101821, @nrfconnect/ncs-eris @nordicjm

cmake/sysbuild/image_signing_installer.cmake
```diff
@@ -51,18 +53,20 @@ function(bm_install_tasks output_hex output_bin)
...
     set(write_block_size 4)
     message(WARNING "slot0_partition write block size devicetree parameter is missing, assuming write block size is 4")
   endif()
 
   sysbuild_get(CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION IMAGE ${DEFAULT_IMAGE} VAR CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION KCONFIG)
   sysbuild_get(CONFIG_ROM_START_OFFSET IMAGE ${DEFAULT_IMAGE} VAR CONFIG_ROM_START_OFFSET KCONFIG)
-  sysbuild_get(CONFIG_FLASH_LOAD_SIZE IMAGE ${DEFAULT_IMAGE} VAR CONFIG_FLASH_LOAD_SIZE KCONFIG)
 #  sysbuild_get(CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION IMAGE ${DEFAULT_IMAGE} VAR CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION KCONFIG)
 
   # Custom TLV 0x0a is set to 0x01 to indicate that this is an installer image
```
PizzaAllTheWay, line 68: Typo: comment says `0x0a` but the code uses `0xa0`. Should be 
`# Custom TLV 0xa0 is set to 0x01 to indicate that this is an installer image`

PizzaAllTheWay (changes requested):

PizzaAllTheWay:
This branch fails to link the firmware_loader (uart_mcumgr) image:
`undefined reference to mcumgr_serial_tx_pkt`.

Reproduce with `SB_CONFIG_BM_FIRMWARE_LOADER_UART_MCUMGR=y` (what the `.uart` twister test sets via `extra_args`).

**Root cause:** `mcumgr_serial_tx_pkt()` lives in Zephyr's `serial_util.c` and is only compiled when `MCUMGR_TRANSPORT_SERIAL_HAS_SMP_OVER_CONSOLE` is `y`. In Zephyr's `Kconfig.common` that symbol sits inside an `if MCUMGR` scope, so it inherits an implicit `depends on MCUMGR`. The bare-metal build never sets `MCUMGR` (it uses `NCS_BM_MCUMGR`), so the symbol can't turn on and the function is never built — hence the undefined reference.

**Suggested fix** in `subsys/mgmt/mcumgr/transport/Kconfig.bm_uart` — add a second definition of the same symbol. Kconfig merges definitions of the same symbol and ORs their dependencies, so this widens the dependency to `MCUMGR || NCS_BM_MCUMGR` and adds a default that only fires for the UART transport:

```diff
+# Re-declare the hidden Zephyr symbol so its deps become MCUMGR || NCS_BM_MCUMGR.
+# Enables it (and compiles mcumgr_serial_tx_pkt() in serial_util.c) for the BM UART transport.
+# Uses default, not select, to avoid the unmet-MCUMGR-dep error.
+config MCUMGR_TRANSPORT_SERIAL_HAS_SMP_OVER_CONSOLE
+	bool
+	default y if MCUMGR_TRANSPORT_BM_UART
+	depends on NCS_BM_MCUMGR
+

 menuconfig MCUMGR_TRANSPORT_BM_UART
	bool "Bare Metal UART MCUmgr SMP transport"
	...
```

eivindj-nordic:
Added to #806.

PR #806, opened 2026-05-28
Manifest update with serial util changes

eivindj-nordic:
test_bm_samples: PR-116
test_sdk_dfu: PR-814


michalek-no:
rebase

michalek-no (approved):

nvlsianpu (approved):

eivindj-nordic:
@nrfconnect/ncs-bm @nrfconnect/ncs-bm-doc @nrfconnect/ncs-ci, @nrfconnect/ncs-co-boards @nrfconnect/ncs-co-build-system Please have a look!

applications/installer/src/main.c
```diff
@@ -198,6 +200,12 @@ int main(void)
 				}
 #endif
 
+				int pad = (CONFIG_DT_FLASH_WRITE_BLOCK_SIZE -
+					  (process_size % CONFIG_DT_FLASH_WRITE_BLOCK_SIZE)) %
+					  CONFIG_DT_FLASH_WRITE_BLOCK_SIZE;
```
eivindj-nordic: @michalek-no This can be simplified, right? (CONFIG_DT_FLASH_WRITE_BLOCK_SIZE - (process_size % CONFIG_DT_FLASH_WRITE_BLOCK_SIZE) will never be larger than CONFIG_DT_FLASH_WRITE_BLOCK_SIZE, so the additional modulo operation is superfluos. 
```suggestion
				int pad = CONFIG_DT_FLASH_WRITE_BLOCK_SIZE -
					  (process_size % CONFIG_DT_FLASH_WRITE_BLOCK_SIZE);
```

michalek-no: that adds 16 if padding isn't needed at all

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -101,6 +105,8 @@ No changes since the latest nRF Connect SDK Bare Metal release.
 Samples
 =======
 
+* The  :kconfig:option:`CONFIG_PSA_CRYPTO` Kconfig option is now used to enable cryptography instead of the :kconfig:option:`CONFIG_NRF_SECURITY` Kconfig option.
```
anhmolt:
```suggestion
* The :kconfig:option:`CONFIG_PSA_CRYPTO` Kconfig option is now used to enable cryptography instead of the :kconfig:option:`CONFIG_NRF_SECURITY` Kconfig option.
```

applications/installer/src/main.c
eivindj-nordic: Right...

samples/bluetooth/ble_hrs/prj.conf
```diff
@@ -6,7 +6,7 @@ CONFIG_LOG_BACKEND_BM_UARTE=y
 CONFIG_SOFTDEVICE=y
 
 # Enable RNG
-CONFIG_NRF_SECURITY=y
+CONFIG_PSA_CRYPTO=y
 CONFIG_MBEDTLS_PSA_CRYPTO_C=y
```
anhmolt:
```suggestion
```
We should remove `CONFIG_MBEDTLS_PSA_CRYPTO_C=y` from the `prj.conf` of samples. The `CONFIG_MBEDTLS_PSA_CRYPTO_C=y` Kconfig is already selected by `CONFIG_PSA_CRYPTO_PROVIDER_CUSTOM` which is the default choice of `PSA_CRYPTO_PROVIDER` in NCS.

subsys/bm_installs/src/bm_installs.c
```diff
@@ -226,7 +226,7 @@ int bm_installs_invalidate(void)
 	}
 
 #if defined(CONFIG_FLASH_HAS_EXPLICIT_ERASE)
-	if ((index_offset + sizeof(bm_installs_data)) >= FIXED_PARTITION_SIZE(
+	if ((index_offset + sizeof(bm_installs_data)) >= PARTITION_SIZE(
 								metadata_partition)) {
```
anhmolt: This can go on one line.

subsys/shell/backends/Kconfig.bm_uarte
```diff
@@ -51,4 +51,12 @@ config SHELL_BACKEND_SERIAL
 config SHELL_MINIMAL
 	default y
 
+# Workaround for added build assert in zephyr that should only be checked when multithreading is
+# enabled. Kconfig is not used with multithreading disabled.
+config SHELL_STACK_SIZE
+	int "Shell thread stack size"
+	default 32
+	help
+	  Stack size for thread created for each instance.
```
anhmolt:
```suggestion
config SHELL_STACK_SIZE
	default 32
```
Is the type/prompt and help needed here?

anhmolt (commented):
One file was overlooked when updating from fixed-partition to mapped-partition:
https://github.com/nrfconnect/sdk-nrf-bm/blob/e6375c4b6a4c8db0fefa8823e064483d960ae12c/boards/nordic/bm_nrf54lv10dk/bm_nrf54lv10dk_nrf54lv10a_cpuapp_s145_softdevice.dts#L24

subsys/shell/backends/Kconfig.bm_uarte
eivindj-nordic: Should not be, removed.

scripts/ci/license_allow_list.yaml
```diff
@@ -61,8 +61,10 @@ Apache-2.0: |
   ^nrf-bm/subsys/mgmt/mcumgr/smp/CMakeLists.txt
   ^nrf-bm/subsys/mgmt/mcumgr/smp/Kconfig
```
nordicjm, line 62: commit missing things like commit ID files were taken from

subsys/mgmt/mcumgr/transport/include/serial.h
```diff
@@ -48,14 +48,6 @@ struct mcumgr_serial_rx_ctxt {
 	/** Length of full packet, as read from header. */
 	uint16_t pkt_len;
 
-#if defined(CONFIG_MCUMGR_TRANSPORT_SERIAL_HAS_SMP_OVER_CONSOLE) && \
-	defined(CONFIG_MCUMGR_TRANSPORT_SERIAL_HAS_RAW_BINARY_NON_SMP_OVER_CONSOLE)
```
nordicjm: so to get this, users must have sdk-nrf-bm which includes sdk-zephyr with this file, none of these parts would be compiled in anyway since the Kconfig would be disabled, but now it takes up more space since they have the files duplicated... So what problem is this trying to fix exactly?

scripts/ci/license_allow_list.yaml
eivindj-nordic: @michalek-no Can you add this?

applications/installer/src/main.c
```diff
@@ -198,6 +198,12 @@ int main(void)
...
```
nordicjm: no idea what these Kconfigs are but they're invalid anyway, get from dt using dt macros

nordicjm (changes requested):

scripts/ci/license_allow_list.yaml
eivindj-nordic: I see the last changes to the code in sdk-zephyr is d5df08a1cc3c14f511951f6ff797ec1e70096f22, which is identical to the copied code, so I'll add that.

applications/installer/src/main.c
eivindj-nordic: Replaced with DT macros.

subsys/mgmt/mcumgr/transport/include/serial.h
eivindj-nordic: Removed the copying of the header.

subsys/mgmt/mcumgr/transport/src/serial_util.c
```diff
@@ -17,10 +17,6 @@
 #include <zephyr/mgmt/mcumgr/smp/smp.h>
 #include <zephyr/mgmt/mcumgr/transport/serial.h>
 
-#if defined(CONFIG_MCUMGR_TRANSPORT_SERIAL_HAS_RAW_BINARY_NON_SMP_OVER_CONSOLE)
```
nordicjm: this is what I meant by why, these parts are #ifdef'd out meaning they are not built, so why is a file from sdk-zephyr being duplicated so users need **both** files for something that would compile exactly the same if it just uses the sdk-zephyr one?

michalek-no: Otherwise we need to drag other Kconfigs with its spaghetti

eivindj-nordic: To elaborate a bit, we require functionality (`mcumgr_serial_tx_pkt`) that is now only available if the `CONFIG_MCUMGR_TRANSPORT_SERIAL_HAS_SMP_OVER_CONSOLE` Kconfig option is set. That again depends on the Zephyr `CONFIG_MCUMGR` Kconfig option that is not enabled with BM. So to get access to the `mcumgr_serial_tx_pkt()` function we cannot only select `CONFIG_MCUMGR_TRANSPORT_SERIAL_HAS_SMP_OVER_CONSOLE` in the BM transport. If that was the case we would probably be ok with using the  Zephyr file (haven't checked), with the risk that we would have to go through this again for the next Zephyr upmerge.

nordicjm:
```
config SOME_BM_CONFIG
        select MCUMGR_TRANSPORT_SERIAL_HAS_SMP_OVER_CONSOLE
```
?

eivindj-nordic: Not without redefining the `MCUMGR_TRANSPORT_SERIAL_HAS_SMP_OVER_CONSOLE` Kconfig as well in BM (See #813), or including the `subsys/mgmt/mcumgr/transport/common.kconfig` file from Zephyr as a resource in BM. So it will be a bit of a hacky solution either way.

nordicjm: if it needs redefining then it needs redefining, it seems pointless to me to duplicate a file then have to maintain said file against the upstream file... who is going to do that to make sure bugs are fixed? How many files need to be checked with each upmerge?

eivindj-nordic: Removed C file and added the `MCUMGR_TRANSPORT_SERIAL_HAS_SMP_OVER_CONSOLE` Kconfig. 
Diff: https://github.com/nrfconnect/sdk-nrf-bm/compare/8f462e28fd4b0a52fa90dff68a713c152d9c46be..5395da78d520dfe4a919c9e3482c2da06b249890

nordicjm (approved):

doc/nrf-bm/libraries/bluetooth/peer_manager.rst
```diff
@@ -453,7 +453,7 @@ LE Secure Connections is an optional functionality of the Peer Manager and is di
 However, if you want to use it, keep in mind that it depends on nRF Security to generate the Diffie-Helman key pair.
```
b-gent: Looking at the Kconfig change, shouldn't this say:
```suggestion
However, if you want to use it, keep in mind that it depends on PSA Crypto to generate the Diffie-Helman key pair.
```

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -33,7 +33,11 @@ No changes since the latest nRF Connect SDK Bare Metal release.
 Boards
 ======
 
-No changes since the latest nRF Connect SDK Bare Metal release.
+* Updated all boards to use ``zephyr,mapped-partition`` instead of ``zephyr,fixed-partitions`` and ``zephyr,fixed-subpartitions``.
+  When referencing partition nodes the code now uses ``PARTITION_*`` macros instead of ``DT_*`` and ``FIXED_PARTITION_*`` macros.
+  The use of the :kconfig:option:`CONFIG_FLASH_LOAD_OFFSET` Kconfig option is also replaced by ``PARTITION_*`` macros.
+
+* Reduced the number of required board qualifiers with one as of changes upstream.
```
b-gent:
```suggestion
* Reduced the number of required board qualifiers by one, as of changes upstream.
```
(upstream Zephyr?)

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -33,7 +33,11 @@ No changes since the latest nRF Connect SDK Bare Metal release.
 Boards
 ======
 
-No changes since the latest nRF Connect SDK Bare Metal release.
+* Updated all boards to use ``zephyr,mapped-partition`` instead of ``zephyr,fixed-partitions`` and ``zephyr,fixed-subpartitions``.
+  When referencing partition nodes the code now uses ``PARTITION_*`` macros instead of ``DT_*`` and ``FIXED_PARTITION_*`` macros.
```
b-gent:
```suggestion
  When referencing partition nodes, the code now uses ``PARTITION_*`` macros instead of ``DT_*`` and ``FIXED_PARTITION_*`` macros.
```

b-gent (approved):

doc/nrf-bm/libraries/bluetooth/peer_manager.rst
eivindj-nordic: Good catch!

doc/nrf-bm/release_notes/release_notes_changelog.rst
eivindj-nordic: Yes

anhmolt (approved):

PR #812, opened 2026-06-23
bluetooth: services: dis: add runtime configuration support

PizzaAllTheWay:
Allow the Device Information Service to be configured at run time through the new `ble_dis_values` structure, passed via the `values` field of `ble_dis_config`. When `values` is `NULL`, the service is built from the Kconfig defaults as before, otherwise it fully defines the characteristics, with any `NULL` or empty field omitting its characteristic. This lets a single precompiled binary serve multiple devices, for example by setting only the serial number at run time.

A `BLE_DIS_VALUES_DEFAULTS` macro is also added to pre-populate a `ble_dis_values` struct from the Kconfig defaults at run time, making it easy to load all defaults and selectively override individual fields before calling `ble_dis_init`.

Also updates the surrounding documentation and tests. Documented the Kconfig options, refreshed the `BLE_DIS` API docs to cover the runtime configuration, and reworked the unit tests into focused cases covering the runtime, partial, single-characteristic, Kconfig default, and macro-based configuration paths.

### DIFF
--- a/doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
+++ b/doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
@@ -34,15 +34,15 @@ You can configure the service using the following Kconfig options that provide t
 * :kconfig:option:`CONFIG_BLE_DIS_HW_REVISION` - Sets the hardware revision.
 * :kconfig:option:`CONFIG_BLE_DIS_FW_REVISION` - Sets the firmware revision.
 * :kconfig:option:`CONFIG_BLE_DIS_SW_REVISION` - Sets the software revision.
-* :kconfig:option:`CONFIG_BLE_DIS_SYSTEM_ID` - Includes the system ID characteristic in the Device Information Service.
-* :kconfig:option:`CONFIG_BLE_DIS_SYSTEM_ID_OUI` - Sets the organization unique ID.
-* :kconfig:option:`CONFIG_BLE_DIS_SYSTEM_ID_MID` - Sets the manufacturer unique ID.
-* :kconfig:option:`CONFIG_BLE_DIS_PNP_ID` - Includes plug and play ID characteristic.
+* :kconfig:option:`CONFIG_BLE_DIS_SYSTEM_ID` - Includes the System ID characteristic in the Device Information Service.
+* :kconfig:option:`CONFIG_BLE_DIS_SYSTEM_ID_OUI` - Sets the Organizationally Unique ID.
+* :kconfig:option:`CONFIG_BLE_DIS_SYSTEM_ID_MID` - Sets the Manufacturer ID.
+* :kconfig:option:`CONFIG_BLE_DIS_PNP_ID` - Includes the Plug and Play ID characteristic.
 * :kconfig:option:`CONFIG_BLE_DIS_PNP_VID_SRC` - Sets the vendor ID source.
 * :kconfig:option:`CONFIG_BLE_DIS_PNP_VID` - Sets the vendor ID.
 * :kconfig:option:`CONFIG_BLE_DIS_PNP_PID` - Sets the product ID.
 * :kconfig:option:`CONFIG_BLE_DIS_PNP_VER` - Sets the product version.
-* :kconfig:option:`CONFIG_BLE_DIS_REGULATORY_CERT` - Includes IEEE regulatory certifications.
+* :kconfig:option:`CONFIG_BLE_DIS_REGULATORY_CERT` - Includes the IEEE 11073-20601 Regulatory Certification characteristic.
 * :kconfig:option:`CONFIG_BLE_DIS_REGULATORY_CERT_LIST` - Sets the regulatory certification list.
 
 Initialization
--- a/doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
+++ b/doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
@@ -15,12 +15,18 @@ Overview
 During initialization, the module adds the Device Information Service to the Bluetooth LE stack database.
 It then encodes the supplied information, and adds the corresponding characteristics.
 
+The characteristic values follow an all-or-nothing model:
+
+* If the application supplies a :c:struct:`ble_dis_values` instance at initialization, the values fully define every characteristic.
+  The Kconfig options are ignored and any member left ``NULL`` (or empty for strings) omits that characteristic.
+* If no run-time values are supplied, all characteristics are built from the :kconfig:option:`CONFIG_BLE_DIS_*` Kconfig build-time defaults.
+
 Configuration
 *************
 
 Set the :kconfig:option:`CONFIG_BLE_DIS` Kconfig option to enable the service.
 
-The DIS service can be configured by using the following Kconfig options:
+You can configure the service using the following Kconfig options that provide the build-time default values:
 
 * :kconfig:option:`CONFIG_BLE_DIS_MANUFACTURER_NAME` - Sets the manufacturer name.
 * :kconfig:option:`CONFIG_BLE_DIS_MODEL_NUMBER` - Sets the model number.
@@ -42,13 +48,16 @@ The DIS service can be configured by using the following Kconfig options:
 Initialization
 ==============
 
-The service is initialized by calling the :c:func:`ble_dis_init` function.
-Configuration is otherwise done through the Kconfig options.
+To initialize the service, call the :c:func:`ble_dis_init` function.
+See the :c:struct:`ble_dis_config` struct for configuration details.
+
+Set the ``values`` field of :c:struct:`ble_dis_config` to a :c:struct:`ble_dis_values` instance to supply run-time values, or to ``NULL`` to use the Kconfig build-time defaults.
 
 Usage
 *****
 
-When enabled, the module will add the Device Information Service with information as specified by the Kconfig options.
+When enabled, the module adds the Device Information Service with the characteristic values supplied at run time.
+If no run-time values are supplied, the values specified by the Kconfig options are used.
 
 Dependencies
 ************
--- a/doc/nrf-bm/release_notes/release_notes_changelog.rst
+++ b/doc/nrf-bm/release_notes/release_notes_changelog.rst
@@ -106,6 +106,11 @@ Bluetooth LE Services
 
    * Changed :c:member:`ble_scan_filter_data.addr_filter.addr` and :c:member:`ble_scan_filter_data.name_filter.name` to ``const`` in the :c:struct:`ble_scan_filter_data` structure.
 
+* :ref:`lib_ble_service_dis`:
+
+   * Added support for configuring the Device Information Service characteristics at run time through the new :c:struct:`ble_dis_values` structure, passed using the ``values`` field of :c:struct:`ble_dis_config`.
+     When ``values`` is ``NULL``, the service is built from the Kconfig defaults as before.
+
 Libraries for NFC
 -----------------
 
--- a/include/bm/bluetooth/services/ble_dis.h
+++ b/include/bm/bluetooth/services/ble_dis.h
@@ -6,7 +6,7 @@
 /** @file
  *
  * @defgroup ble_dis Device Information Service
- *
+ * @{
  * @brief Device Information Service module.
  *
  * @details This module implements the Device Information Service.
@@ -19,6 +19,7 @@
 #define BLE_DIS_H__
 
 #include <stdint.h>
+#include <zephyr/sys/util.h>
 #include <bm/bluetooth/ble_common.h>
 
 #ifdef __cplusplus
@@ -31,9 +32,97 @@ extern "C" {
 		.device_info_char.read = BLE_GAP_CONN_SEC_MODE_OPEN,                               \
 	}
 
+/** @brief Default characteristic values, populated from the @c CONFIG_BLE_DIS_* Kconfig options. */
+#define BLE_DIS_VALUES_DEFAULTS                                                                    \
+	{                                                                                          \
+		.manufacturer_name = CONFIG_BLE_DIS_MANUFACTURER_NAME,                             \
+		.model_number      = CONFIG_BLE_DIS_MODEL_NUMBER,                                  \
+		.serial_number     = CONFIG_BLE_DIS_SERIAL_NUMBER,                                 \
+		.hw_revision       = CONFIG_BLE_DIS_HW_REVISION,                                   \
+		.fw_revision       = CONFIG_BLE_DIS_FW_REVISION,                                   \
+		.sw_revision       = CONFIG_BLE_DIS_SW_REVISION,                                   \
+		IF_ENABLED(CONFIG_BLE_DIS_SYSTEM_ID, (                                             \
+			.sys_id = &(const struct ble_dis_sys_id){                                  \
+				.manufacturer_id            = CONFIG_BLE_DIS_SYSTEM_ID_MID,        \
+				.organizationally_unique_id = CONFIG_BLE_DIS_SYSTEM_ID_OUI,        \
+			},))                                                                       \
+		IF_ENABLED(CONFIG_BLE_DIS_PNP_ID, (                                                \
+			.pnp_id = &(const struct ble_dis_pnp_id){                                  \
+				.vendor_id_source = CONFIG_BLE_DIS_PNP_VID_SRC,                    \
+				.vendor_id        = CONFIG_BLE_DIS_PNP_VID,                        \
+				.product_id       = CONFIG_BLE_DIS_PNP_PID,                        \
+				.product_version  = CONFIG_BLE_DIS_PNP_VER,                        \
+			},))                                                                       \
+		IF_ENABLED(CONFIG_BLE_DIS_REGULATORY_CERT, (                                       \
+			.reg_cert = &(const struct ble_dis_reg_cert){                              \
+				.list = (const uint8_t[8])                                         \
+					sys_uint64_to_array(CONFIG_BLE_DIS_REGULATORY_CERT_LIST),  \
+				.len  = 8,                                                         \
+			},))                                                                       \
+	}
+
 /**
- * @brief Device Information service configuration.
+ * @name PnP ID Vendor ID Source values
+ * @{
  */
+/** Vendor ID assigned by the Bluetooth SIG. */
+#define BLE_DIS_VENDOR_ID_SRC_BLUETOOTH_SIG 1
+/** Vendor ID assigned by the USB Implementer's Forum. */
+#define BLE_DIS_VENDOR_ID_SRC_USB_IMPL_FORUM 2
+/** @} */
+
+/** @brief System ID characteristic value. */
+struct ble_dis_sys_id {
+	/** Manufacturer ID. Only the 5 least significant octets are used. */
+	uint64_t manufacturer_id;
+	/** Organizationally unique ID. Only the 3 least significant octets are used. */
+	uint32_t organizationally_unique_id;
+};
+
+/** @brief PnP ID characteristic value. */
+struct ble_dis_pnp_id {
+	/** Vendor ID source. See @ref BLE_DIS_VENDOR_ID_SRC_BLUETOOTH_SIG. */
+	uint8_t vendor_id_source;
+	/** Vendor ID. */
+	uint16_t vendor_id;
+	/** Product ID. */
+	uint16_t product_id;
+	/** Product version. */
+	uint16_t product_version;
+};
+
+/** @brief IEEE 11073-20601 Regulatory Certification Data List characteristic value. */
+struct ble_dis_reg_cert {
+	/** Encoded opaque structure based on the IEEE 11073-20601 specification. */
+	const uint8_t *list;
+	/** Length of @ref ble_dis_reg_cert.list. */
+	uint16_t len;
+};
+
+/** @brief Device Information Service characteristic values. */
+struct ble_dis_values {
+	/** Manufacturer Name String. @c NULL or empty string omits the characteristic. */
+	const char *manufacturer_name;
+	/** Model Number String. @c NULL or empty string omits the characteristic. */
+	const char *model_number;
+	/** Serial Number String. @c NULL or empty string omits the characteristic. */
+	const char *serial_number;
+	/** Hardware Revision String. @c NULL or empty string omits the characteristic. */
+	const char *hw_revision;
+	/** Firmware Revision String. @c NULL or empty string omits the characteristic. */
+	const char *fw_revision;
+	/** Software Revision String. @c NULL or empty string omits the characteristic. */
+	const char *sw_revision;
+
+	/** System ID. @c NULL omits the characteristic. */
+	const struct ble_dis_sys_id *sys_id;
+	/** Plug and Play ID. @c NULL omits the characteristic. */
+	const struct ble_dis_pnp_id *pnp_id;
+	/** IEEE 11073-20601 Regulatory Certification Data List. @c NULL omits characteristic. */
+	const struct ble_dis_reg_cert *reg_cert;
+};
+
+/** @brief Device Information Service configuration. */
 struct ble_dis_config {
 	/** Security configuration. */
 	struct {
@@ -45,19 +134,29 @@ struct ble_dis_config {
 			ble_gap_conn_sec_mode_t read;
 		} device_info_char;
 	} sec_mode;
+
+	/** Characteristic values. @c NULL builds the service from the Kconfig defaults. */
+	const struct ble_dis_values *values;
 };
 
 /**
  * @brief Initialize the Device Information Service.
  *
  * @details This call allows the application to initialize the device information service.
- *          It adds the DIS service and DIS characteristics to the database, using the
- *          values supplied through the Kconfig options.
+ *          It adds the DIS service and DIS characteristics to the database.
  *
- * @param dis_config Device Information Service configuration.
+ * @param[in] dis_config Device Information Service configuration.
+ *                       If @ref ble_dis_config::values is @c NULL the values are set based on
+ *                       the @c CONFIG_BLE_DIS_* Kconfig options.
+ *                       If a member of @ref ble_dis_values is left @c NULL or empty string
+ *                       the characteristic is omitted.
  *
- * @return NRF_SUCCESS on successful initialization of service.
+ * @retval NRF_SUCCESS On success.
  * @retval NRF_ERROR_NULL If @p dis_config is @c NULL.
+ * @return In addition, this function may return any error
+ *         returned by the following SoftDevice functions:
+ *         - @ref sd_ble_gatts_service_add()
+ *         - @ref sd_ble_gatts_characteristic_add()
  */
 uint32_t ble_dis_init(const struct ble_dis_config *dis_config);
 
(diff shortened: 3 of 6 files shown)


doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
```diff
@@ -15,40 +15,45 @@ Overview
 During initialization, the module adds the Device Information Service to the Bluetooth LE stack database.
 It then encodes the supplied information, and adds the corresponding characteristics.
 
+The characteristic values are taken either from run-time values supplied by the application or from the Kconfig options.
```
eivindj-nordic: Might be formulated better than this, but I think it would be good to give some brief overview of when Kconfigs are used and not. 
```suggestion
The characteristic values can be configured during initialization of the service. If the values are not provided the defaults set by the Kconfig options are used.
```

subsys/bluetooth/services/ble_dis/dis.c
```diff
@@ -83,27 +117,75 @@ uint32_t ble_dis_init(const struct ble_dis_config *dis_config)
...
+	 * otherwise fall back to the Kconfig defaults.
+	 */
+	uint8_t sys_id_buf[SYS_ID_LEN];
+	uint8_t pnp_id_buf[PNP_ID_LEN];
+	const uint8_t *sys_id_val = NULL;
+	const uint8_t *pnp_id_val = NULL;
+	const uint8_t *reg_cert_val = NULL;
+	uint16_t sys_id_len = 0;
+	uint16_t pnp_id_len = 0;
+	uint16_t reg_cert_len = 0;
```
eivindj-nordic: Move to start of the function.

subsys/bluetooth/services/ble_dis/dis.c
```diff
@@ -83,27 +117,75 @@ uint32_t ble_dis_init(const struct ble_dis_config *dis_config)
...
+
+	if (vals) {
```
eivindj-nordic, line 104: Do these need to be all or nothing? 
I think it should be possible to overwrite one of the Kconfigs without overwriting all. This also needs to be mirrored in the documentation/comments. Alternatively we are fine with not keeping the backward compatibility, and we should deprecate the Kconfig options. Discuss with  @lemrey or @anhmolt if I'm not here.

subsys/bluetooth/services/ble_dis/Kconfig
```diff
@@ -4,100 +4,134 @@
...
+	help
+	  Enable the Bluetooth LE Device Information Service (DIS).
+	  The values configured here are used as the build-time defaults,
+	  applied when the application passes no run-time values to ble_dis_init().
 
 if BLE_DIS
 
 config BLE_DIS_MANUFACTURER_NAME
-	string "Manufacturer name"
+	string "Manufacturer Name String"
```
eivindj-nordic: string is already in the prompt type, should not be required. Here and others.

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
```diff
@@ -15,40 +15,45 @@ Overview
...
+* :kconfig:option:`CONFIG_BLE_DIS_SYSTEM_ID` - Includes the System ID characteristic in the Device Information Service.
+* :kconfig:option:`CONFIG_BLE_DIS_SYSTEM_ID_OUI` - Sets the Organizationally Unique ID.
+* :kconfig:option:`CONFIG_BLE_DIS_SYSTEM_ID_MID` - Sets the Manufacturer ID.
+* :kconfig:option:`CONFIG_BLE_DIS_PNP_ID` - Includes the Plug and Play ID characteristic.
 * :kconfig:option:`CONFIG_BLE_DIS_PNP_VID_SRC` - Sets the vendor ID source.
 * :kconfig:option:`CONFIG_BLE_DIS_PNP_VID` - Sets the vendor ID.
 * :kconfig:option:`CONFIG_BLE_DIS_PNP_PID` - Sets the product ID.
 * :kconfig:option:`CONFIG_BLE_DIS_PNP_VER` - Sets the product version.
-* :kconfig:option:`CONFIG_BLE_DIS_REGULATORY_CERT` - Includes IEEE regulatory certifications.
+* :kconfig:option:`CONFIG_BLE_DIS_REGULATORY_CERT` - Includes the IEEE 11073-20601 Regulatory Certification characteristic.
```
eivindj-nordic, line 45: This does not belong in this PR, at least not in the commit introducing the runtime configuration.

subsys/bluetooth/services/ble_dis/dis.c
```diff
@@ -26,6 +27,14 @@
 			    (.len = __VA_ARGS__,))                                                 \
 	}
 
+/* Resolve a string characteristic value:
+ * when run-time values are supplied,
+ * use the field (a NULL/empty field omits the characteristic via a "" -> zero-length value),
+ * otherwise fall back to the Kconfig default.
+ */
+#define resolve_str(_cfg_val, _kconfig_val)                                                        \
+	(vals ? ((_cfg_val) != NULL ? (_cfg_val) : "") : (_kconfig_val))
```
anhmolt: What is `vals` in this macro? I would expect it to be in the parameter list.

include/bm/bluetooth/services/ble_dis.h
```diff
@@ -45,19 +104,31 @@ struct ble_dis_config {
...
+ * @note The read security requirement is always taken from @p dis_config->sec_mode.
  *
- * @param dis_config Device Information Service configuration.
+ * @param[in] dis_config Device Information Service configuration.
  *
- * @return NRF_SUCCESS on successful initialization of service.
+ * @retval NRF_SUCCESS On successful initialization of the service.
  * @retval NRF_ERROR_NULL If @p dis_config is @c NULL.
+ * @return If functions from other modules return errors to this function,
+ *         the @ref nrf_error are propagated.
```
anhmolt:
```Suggestion
 * @return In addition, this function may return any error
 *         returned by the following SoftDevice functions:
 *         - @ref sd_ble_gatts_service_add()
 *         - @ref sd_ble_gatts_characteristic_add()
```

include/bm/bluetooth/services/ble_dis.h
```diff
@@ -45,19 +104,31 @@ struct ble_dis_config {
...
+ *          Otherwise @p dis_config->values fully defines them and the Kconfig defaults are ignored,
+ *          any member left @c NULL (or empty, for strings) omits its characteristic.
+ *
+ * @note The read security requirement is always taken from @p dis_config->sec_mode.
  *
- * @param dis_config Device Information Service configuration.
+ * @param[in] dis_config Device Information Service configuration.
  *
- * @return NRF_SUCCESS on successful initialization of service.
+ * @retval NRF_SUCCESS On successful initialization of the service.
```
anhmolt:
```suggestion
 * @retval NRF_SUCCESS On success.
```
Not necessary to re-iterate on what the function does.

include/bm/bluetooth/services/ble_dis.h
```diff
@@ -45,19 +104,31 @@ struct ble_dis_config {
...
+ *          The characteristic values follow an "all or nothing" model:
+ *          when @p dis_config->values is @c NULL,
+ *          all values are taken from the @c CONFIG_BLE_DIS_* Kconfig defaults.
+ *          Otherwise @p dis_config->values fully defines them and the Kconfig defaults are ignored,
+ *          any member left @c NULL (or empty, for strings) omits its characteristic.
+ *
+ * @note The read security requirement is always taken from @p dis_config->sec_mode.
  *
- * @param dis_config Device Information Service configuration.
+ * @param[in] dis_config Device Information Service configuration.
```
anhmolt, line 148:
```suggestion
 * @param[in] dis_config Device Information Service configuration.
 *                       If @ref ble_dis_config::values is @c NULL the values are set based on
 *                       the @c CONFIG_BLE_DIS_* Kconfig options.
 *                       If a member of @ref ble_dis_values is left @c NULL (or empty, for strings)
 *                       the characteristic is omitted.
```

include/bm/bluetooth/services/ble_dis.h
```diff
@@ -45,19 +104,31 @@ struct ble_dis_config {
...
  * @details This call allows the application to initialize the device information service.
- *          It adds the DIS service and DIS characteristics to the database, using the
- *          values supplied through the Kconfig options.
+ *          It adds the DIS service and DIS characteristics to the database.
+ *
+ *          The characteristic values follow an "all or nothing" model:
+ *          when @p dis_config->values is @c NULL,
+ *          all values are taken from the @c CONFIG_BLE_DIS_* Kconfig defaults.
+ *          Otherwise @p dis_config->values fully defines them and the Kconfig defaults are ignored,
+ *          any member left @c NULL (or empty, for strings) omits its characteristic.
```
anhmolt: I think some of this can be removed and some moved to the `params[in] dis_config`. See one of my other comments.

include/bm/bluetooth/services/ble_dis.h
```diff
@@ -45,19 +104,31 @@ struct ble_dis_config {
...
+ *
+ * @note The read security requirement is always taken from @p dis_config->sec_mode.
```
anhmolt: Not sure this is needed.

tests/unit/subsys/bluetooth/services/ble_dis/src/unity_test.c
```diff
@@ -15,179 +15,529 @@
...
 }
 
-void test_ble_dis_init_error_invalid_param(void)
+void test_ble_dis_init_error_service_add(void)
 {
+	/* sd_ble_gatts_service_add() fails, error propagated. */
 	uint32_t nrf_err;
 
-	__cmock_sd_ble_gatts_service_add_Stub(stub_sd_ble_gatts_service_add_invalid_param);
+	__cmock_sd_ble_gatts_service_add_Stub(stub_service_add_error);
```
anhmolt: Is a stub needed here? Check if `__cmock_sd_ble_gatts_service_add_ExpectAnyArgsAndReturn(NRF_ERROR_INVALID_PARAM);` is sufficient.

tests/unit/subsys/bluetooth/services/ble_dis/src/unity_test.c
```diff
@@ -15,179 +15,529 @@
...
+void test_ble_dis_init_error_char_add(void)
+{
+	/* sd_ble_gatts_characteristic_add() fails, error propagated. */
+	uint32_t nrf_err;
 
-	__cmock_sd_ble_gatts_service_add_Stub(stub_sd_ble_gatts_service_add);
-	__cmock_sd_ble_gatts_characteristic_add_Stub(
-		stub_sd_ble_gatts_characteristic_add_invalid_param);
+	__cmock_sd_ble_gatts_service_add_Stub(stub_service_add);
+	__cmock_sd_ble_gatts_characteristic_add_Stub(stub_char_add_error);
```
anhmolt: Same here. Is a stub really needed?

nordicjm (approved):

subsys/bluetooth/services/ble_dis/dis.c
```diff
@@ -58,6 +67,28 @@ static const uint8_t regulatory_certifications[IEEE_CERT_LEN] =
 
 LOG_MODULE_REGISTER(ble_dis, CONFIG_BLE_DIS_LOG_LEVEL);
 
+/* Encode a run-time System ID into its on-air representation. */
+static void sys_id_encode(uint8_t *buf, const struct ble_dis_sys_id *id)
+{
+	buf[0] = (id->manufacturer_id & 0x00000000ff);
```
lemrey: Could you have used the `sys_put_` functions here, so that you encode these fields in the same way?

subsys/bluetooth/services/ble_dis/dis.c
lemrey: Today, whether the SYS ID, PNP ID & Regularatory Cert characteristics are available in the service is controlled by their relative Kconfig option. In the current implementation they are the only ones that require a static buffer because they are encoded and not simple strings, so putting them behind a Kconfig helps to shave off a few bytes of RAM, in case they are not used.

With this patch, it will no longer be the case that those characteristics are added only if their relative Kconfig option is set, but also when the user sets them up at runtime.

This is confusing, because users may expect to have to enable the Kconfig option to add that characteristic also when doing manual initialization, but that's not the case. So they turn that option on, pass some value during initialization and everything works, but the implementation now has allocated the buffer for those characteristics twice, once statically (unused), and once on the stack.

In practice, allowing runtime initialization of these characteristics defeats the goal of saving a few bytes of RAM in case they are not used, since they are always allocated on the stack regardless, so the Kconfig are basically useless.

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
PizzaAllTheWay: fixed but might need to edit again depending on the comments left by @lemrey and behaviour changes

subsys/bluetooth/services/ble_dis/dis.c
PizzaAllTheWay: fixed

subsys/bluetooth/services/ble_dis/Kconfig
PizzaAllTheWay: fixed

subsys/bluetooth/services/ble_dis/dis.c
PizzaAllTheWay: Yeah will have a chat with @lemrey when he is available about this

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
PizzaAllTheWay: fixed, made own commit for this

subsys/bluetooth/services/ble_dis/dis.c
PizzaAllTheWay: Very good point. `vals` wasn't a parameter, the macro implicitly used the `vals` local from `ble_dis_init` scope. Bad design on my part basically. Fixed by making `vals` an explicit parameter, passed in at every call site now.

include/bm/bluetooth/services/ble_dis.h
PizzaAllTheWay: fixed

include/bm/bluetooth/services/ble_dis.h
PizzaAllTheWay: fixed

include/bm/bluetooth/services/ble_dis.h
PizzaAllTheWay: fixed

include/bm/bluetooth/services/ble_dis.h
PizzaAllTheWay: fixed

include/bm/bluetooth/services/ble_dis.h
PizzaAllTheWay: fixed

tests/unit/subsys/bluetooth/services/ble_dis/src/unity_test.c
PizzaAllTheWay: Good point, fixed

tests/unit/subsys/bluetooth/services/ble_dis/src/unity_test.c
PizzaAllTheWay: fixed

subsys/bluetooth/services/ble_dis/dis.c
PizzaAllTheWay: Good point, fixed

subsys/bluetooth/services/ble_dis/dis.c
PizzaAllTheWay: Let's discuss this when you are in the office

subsys/bluetooth/services/ble_dis/dis.c
PizzaAllTheWay: Discussed with @lemrey, we're keeping the all or nothing model. For partial overrides, the new `BLE_DIS_VALUES_DEFAULTS` macro lets you load all Kconfig defaults at runtime and then selectively override individual fields before calling `ble_dis_init()`

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
PizzaAllTheWay: edited it to align with the API

subsys/bluetooth/services/ble_dis/dis.c
PizzaAllTheWay: Discussed in the office. The static pre-encoded arrays have been removed, the only allocation now is the two small stack buffers (`sys_id_buf` 8 Bytes and `pnp_id_buf` 7 Bytes), which are a fixed and acceptable cost per `ble_dis_init()` call.

For the Kconfig only path (`vals == NULL`) the `CONFIG_BLE_DIS_SYSTEM_ID` / `CONFIG_BLE_DIS_PNP_ID` / `CONFIG_BLE_DIS_REGULATORY_CERT` guards still control whether those characteristics are added, so the options remain meaningful there. For the runtime path (`vals != NULL`) inclusion is controlled purely by whether the pointer fields are `NULL`, and the Kconfig options have no effect, which is the intended all or nothing behaviour.

subsys/bluetooth/services/ble_dis/dis.c
```diff
@@ -83,27 +91,86 @@ uint32_t ble_dis_init(const struct ble_dis_config *dis_config)
...
+			.product_version  = CONFIG_BLE_DIS_PNP_VER,
+		};
+		pnp_id_encode(pnp_id_buf, &kconfig_pnp_id);
+		pnp_id_val = pnp_id_buf;
+		pnp_id_len = sizeof(pnp_id_buf);
 #endif
 #if CONFIG_BLE_DIS_REGULATORY_CERT
-		gatt_char(BLE_UUID_IEEE_REGULATORY_CERTIFICATION_DATA_LIST_CHAR,
-			  regulatory_certifications, sizeof(regulatory_certifications)),
+		static const uint8_t kconfig_reg_cert[IEEE_CERT_LEN] =
```
lemrey: No need for these buffers to be `static`

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -106,6 +106,13 @@ Bluetooth LE Services
 
    * Changed :c:member:`ble_scan_filter_data.addr_filter.addr` and :c:member:`ble_scan_filter_data.name_filter.name` to ``const`` in the :c:struct:`ble_scan_filter_data` structure.
 
+* :ref:`lib_ble_service_dis`:
+
+   * Added:
+
+      * Support for configuring the Device Information Service characteristics at run time through the new :c:struct:`ble_dis_values` structure, passed using the ``values`` field of :c:struct:`ble_dis_config`.
+        When ``values`` is ``NULL``, the service is built from the Kconfig defaults as before.
```
```suggestion
   * Added support for configuring the Device Information Service characteristics at run time through the new :c:struct:`ble_dis_values` structure, passed using the ``values`` field of :c:struct:`ble_dis_config`.
     When ``values`` is ``NULL``, the service is built from the Kconfig defaults as before.
```

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
```diff
@@ -15,40 +15,49 @@ Overview
...
 * :kconfig:option:`CONFIG_BLE_DIS_REGULATORY_CERT_LIST` - Sets the regulatory certification list.
 
 Initialization
 ==============
 
 The service is initialized by calling the :c:func:`ble_dis_init` function.
```
peknis:
```suggestion
To initialize the service, call the :c:func:`ble_dis_init` function.
```

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
```diff
@@ -15,40 +15,49 @@ Overview
...
+  The Kconfig options are ignored, and any member left ``NULL`` (or empty for strings) omits that characteristic.
+* If no run-time values are supplied, all characteristics are built from the :kconfig:option:`CONFIG_BLE_DIS_*` Kconfig build-time defaults.
+
 Configuration
 *************
 
 Set the :kconfig:option:`CONFIG_BLE_DIS` Kconfig option to enable the service.
 
-The DIS service can be configured by using the following Kconfig options:
+The service can be configured using the following Kconfig options, which provide the build-time default values:
```
peknis:
```suggestion
You can configure the service using the following Kconfig options that provide the build-time default values:
```

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
```diff
@@ -15,40 +15,49 @@ Overview
 During initialization, the module adds the Device Information Service to the Bluetooth LE stack database.
 It then encodes the supplied information, and adds the corresponding characteristics.
 
+The characteristic values follow an all-or-nothing model:
+
+* If the application supplies a :c:struct:`ble_dis_values` instance at initialization, those values fully define every characteristic.
```
peknis:
```suggestion
* If the application supplies a :c:struct:`ble_dis_values` instance at initialization, the values fully define every characteristic.
```

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
```diff
@@ -15,40 +15,49 @@ Overview
...
+  The Kconfig options are ignored, and any member left ``NULL`` (or empty for strings) omits that characteristic.
```
peknis:
```suggestion
  The Kconfig options are ignored and any member left ``NULL`` (or empty for strings) omits that characteristic.
```

peknis (dismissed):

doc/nrf-bm/release_notes/release_notes_changelog.rst
PizzaAllTheWay: fixed

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
PizzaAllTheWay: fixed

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
PizzaAllTheWay: fixed

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
PizzaAllTheWay: fixed

doc/nrf-bm/libraries/bluetooth/services/ble_dis.rst
PizzaAllTheWay: fixed

include/bm/bluetooth/services/ble_dis.h
```diff
@@ -45,19 +134,30 @@ struct ble_dis_config {
...
  * @brief Initialize the Device Information Service.
  *
  * @details This call allows the application to initialize the device information service.
- *          It adds the DIS service and DIS characteristics to the database, using the
- *          values supplied through the Kconfig options.
+ *          It adds the DIS service and DIS characteristics to the database.
  *
- * @param dis_config Device Information Service configuration.
+ * @param[in] dis_config Device Information Service configuration.
+ *                       If @p dis_config is @c NULL the function returns @ref NRF_ERROR_NULL.
```
lemrey: I think this line can be omitted, since it's redundant with 
` * @retval NRF_ERROR_NULL If @p dis_config is @c NULL.` below

tests/unit/subsys/bluetooth/services/ble_dis/src/unity_test.c
```diff
@@ -15,179 +15,588 @@
...
+
+	__cmock_sd_ble_gatts_service_add_Stub(stub_service_add);
+	__cmock_sd_ble_gatts_characteristic_add_Stub(stub_char_add);
+
+	nrf_err = ble_dis_init(&config);
+	TEST_ASSERT_EQUAL(NRF_SUCCESS, nrf_err);
+	TEST_ASSERT_EQUAL(1, char_add_call_count);
+}
+
+void test_ble_dis_init_runtime_only_sw_revision(void)
```
lemrey, line 520: Generally these tests are okay but you should extend them to test that the given characteristic was indeed added (and not another one for example).

You could for example count the number of time each given characteristic is added, instead of just asserting on the total.

subsys/bluetooth/services/ble_dis/dis.c
PizzaAllTheWay: fixed

include/bm/bluetooth/services/ble_dis.h
PizzaAllTheWay: fixed

tests/unit/subsys/bluetooth/services/ble_dis/src/unity_test.c
PizzaAllTheWay: fixed

tests/unit/subsys/bluetooth/services/ble_dis/src/unity_test.c
```diff
@@ -15,179 +15,643 @@
...
+ * and that every other characteristic was not, eks:
+ *         ASSERT_ONLY_CHARS_ADDED(.manufacturer_name = true, .serial_number = true);
+ * asserts manufacturer_name and serial_number were added and every other characteristic was not
+ * (unlisted fields default to false).
+ */
+#define ASSERT_ONLY_CHARS_ADDED(...)                                                               \
+	do {                                                                                       \
+		const struct dis_chars_added _expected = { __VA_ARGS__ };                          \
+		TEST_ASSERT_EQUAL_MEMORY(&_expected, &chars_added, sizeof(_expected));             \
+	} while (0)
```
lemrey: Let's not wrap macros that belong to the test framework into other macros. It is preferrable to write
```
TEST_ASSERT_EQUAL(x, y);
TEST_ASSERT_EQUAL(x, y);
TEST_ASSERT_EQUAL(x, y);
```

Than define a local macro to do that. If every test suite implemented local macros to assert, all  assertions would look different in each test suite; that wouldn't help.

PizzaAllTheWay: ahhh oki, makes sense.

PizzaAllTheWay: fixed

PizzaAllTheWay:
@peknis Could you review the documentation again and see that I made correct changes?

lemrey (approved):

include/bm/bluetooth/services/ble_dis.h
```diff
@@ -31,9 +32,97 @@ extern "C" {
...
+struct ble_dis_reg_cert {
+	/** Encoded opaque structure based on the IEEE 11073-20601 specification. */
+	const uint8_t *list;
+	/** Length of @ref ble_dis_reg_cert.list. */
+	uint16_t len;
+};
+
+/** @brief Device Information Service characteristic values. */
+struct ble_dis_values {
+	/** Manufacturer Name String. @c NULL or empty omits the characteristic. */
```
eivindj-nordic: Perhaps `NULL or empty string` is clearer?

eivindj-nordic (approved):

include/bm/bluetooth/services/ble_dis.h
PizzaAllTheWay: Fixed for this and others in the header file

b-gent (approved):

PR #816, opened 2026-07-22
samples: fix uarte RX buffer indexing

eivindj-nordic:
The RX buffer were declared in the wrong order for the ble_nus and ble_nus_central samples causing the uarte to be
provided a buffer with the wrong address offset when the index is 1. This commit fixes that bug and alignes all the samples using uarte with the same pattern.

### DIFF
--- a/doc/nrf-bm/release_notes/release_notes_changelog.rst
+++ b/doc/nrf-bm/release_notes/release_notes_changelog.rst
@@ -172,10 +172,6 @@ Bluetooth LE samples
    * Removed redundant logging of authentication status from the main source file.
      Authentication status is logged by :ref:`lib_peer_manager` library.
 
-* :ref:`ble_nus_central_sample` sample:
-
-   * Fixed the disconnect button handler to only disconnect on button press, and not on button release.
-
 * :ref:`ble_hrs_central_sample` sample:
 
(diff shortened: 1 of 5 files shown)


doc/nrf-bm/release_notes/release_notes_changelog.rst
PizzaAllTheWay, line 204: Add missing changelog entry for the UART RX buffer bug fix in peripheral samples `uarte` and `lpuarte`.

PizzaAllTheWay (commented):
LGTM

PizzaAllTheWay (approved):

doc/nrf-bm/release_notes/release_notes_changelog.rst
eivindj-nordic: I added changelog entries for the NUS samples only, since they were the ones where this is fixing a bug. For the peripheral samples I don't think it is needed since it is only internal rewrite.

PizzaAllTheWay: oki

lemrey (approved):

anhmolt (approved):

b-gent (approved):

PR #820, opened 2026-07-27
scripts: parse RRAM/SRAM/KMU sizes from DTS

PizzaAllTheWay:
Replace the hardcoded per-SoC RRAM/SRAM size tables and hardcoded SRAM base address/KMU reserved area size with values parsed directly from each board's `bm_<name_dk>_<name_soc>_cpuapp_common.dtsi` via `new get_soc_mem_field()` helper, so new SoCs and layout changes no longer need hand editing this script.

### DIFF
--- a/doc/_scripts/gen_memory_layouts.py
+++ b/doc/_scripts/gen_memory_layouts.py
@@ -19,6 +19,8 @@ from __future__ import annotations
 import argparse
 import re
 from dataclasses import dataclass, field
+from enum import Enum
+from functools import cache
 from pathlib import Path
 
 NRF_BM_BASE = Path(__file__).resolve().parents[2]
@@ -134,6 +136,118 @@ def _parse_retained_mem(text: str) -> Partition | None:
 
 
 # ---------------------------------------------------------------------------
+# Universal per-SoC memory-field lookup (parsed straight from DTS)
+# ---------------------------------------------------------------------------
+
+
+class MemField(Enum):
+    """Fields queryable via get_soc_mem_field().
+
+    Each member corresponds to one value read from a SoC's DTS,
+    covering RRAM, SRAM, and the KMU reserved area.
+    """
+
+    RRAM_TOTAL = "rram_total"
+    RRAM_BASE_ADDR = "rram_base_addr"
+    SRAM_TOTAL = "sram_total"
+    SRAM_BASE_ADDR = "sram_base_addr"
+    KMU_SIZE = "kmu_size"
+    KMU_BASE_ADDR = "kmu_base_addr"
+
+
+_REG_RE = re.compile(r"reg\s*=\s*<\s*(0x[0-9a-fA-F]+)\s+(.*?)\s*>")
+_NODE_START_TEMPLATE = r"(?:&{label}\b|\b{label}:\s*\w+@[0-9a-fA-F]+)\s*\{{"
+
+
+def _matching_brace(text: str, open_idx: int) -> int:
+    """Find the closing brace matching an opening one.
+
+    Walks forward from open_idx counting nested braces,
+    and returns the index of the '}' that closes it.
+    """
+    depth = 1
+    i = open_idx + 1
+    while depth and i < len(text):
+        if text[i] == "{":
+            depth += 1
+        elif text[i] == "}":
+            depth -= 1
+        i += 1
+    return i - 1
+
+
+def _find_node_body(text: str, label: str) -> str | None:
+    """Return the body of the last node matching label.
+
+    Handles both override syntax, ``&label { ... };``,
+    and definition syntax, ``label: type@ADDR { ... };``.
+    Order in the file does not matter,
+    since the last occurrence wins.
+    """
+    pattern = re.compile(_NODE_START_TEMPLATE.format(label=label))
+    bodies = [text[m.end() : _matching_brace(text, m.end() - 1)] for m in pattern.finditer(text)]
+    return bodies[-1] if bodies else None
+
+
+def _node_reg(text: str, label: str) -> tuple[int, int]:
+    """Return the (base_addr, size) from label's own reg.
+
+    Ignores any reg that belongs to a nested child node,
+    such as a fixed-partitions child.
+    """
+    body = _find_node_body(text, label)
+    if body is None:
+        raise ValueError(f"Node '{label}' not found in DTS")
+    top_level = body.split("{", 1)[0]  # text before any nested child node
+    m = _REG_RE.search(top_level)
+    if not m:
+        raise ValueError(f"Node '{label}' has no top-level reg")
+    return int(m.group(1), 16), _eval_size_expr(m.group(2))
+
+
+@cache
+def _find_common_dtsi(soc: str, dts_path: Path) -> Path:
+    """Return the path to a SoC's common.dtsi file within dts_path."""
+    matches = sorted(dts_path.glob(f"*_{soc}_cpuapp_common.dtsi"))
+    if not matches:
+        raise FileNotFoundError(f"No *_{soc}_cpuapp_common.dtsi found under {dts_path}")
+    return matches[0]
+
+
+@cache
+def _parse_soc_mem_fields(soc: str, dts_path: Path) -> dict[MemField, int]:
+    """Parse every MemField value for one SoC.
+
+    Reads the SoC's common.dtsi once,
+    and returns all fields as a dict, keyed by MemField.
+    """
+    text = _find_common_dtsi(soc, dts_path).read_text()
+
+    kmu_addr, kmu_size = _node_reg(text, "nrf_kmu_reserved_push_area")
+    rram_addr, rram_size = _node_reg(text, "cpuapp_rram")
+    sram_addr, sram_size = _node_reg(text, "cpuapp_sram")
+
+    return {
+        MemField.KMU_BASE_ADDR: kmu_addr,
+        MemField.KMU_SIZE: kmu_size,
+        MemField.SRAM_BASE_ADDR: kmu_addr,
+        MemField.SRAM_TOTAL: (sram_addr + sram_size) - kmu_addr,
+        MemField.RRAM_BASE_ADDR: rram_addr,
+        MemField.RRAM_TOTAL: rram_size,
+    }
+
+
+def get_soc_mem_field(soc: str, field: MemField, dts_path: Path) -> int:
+    """Look up one memory field for soc, parsed straight from its DTS.
+
+    Example:
+        rram_total = get_soc_mem_field("nrf54l05", MemField.RRAM_TOTAL, dts_path)
+        sram_base  = get_soc_mem_field("nrf54l05", MemField.SRAM_BASE_ADDR, dts_path)
+    """
+    return _parse_soc_mem_fields(soc, dts_path.parent)[field]
+
+
+# ---------------------------------------------------------------------------
 # DTS → BoardConfig
 # ---------------------------------------------------------------------------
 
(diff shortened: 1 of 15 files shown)


doc/_scripts/gen_memory_layouts.py
```diff
@@ -234,7 +331,8 @@ def parse_dts(dts_path: Path, common_text: str = "") -> BoardConfig:
     rram_parts.sort(key=lambda p: p.offset)
 
     # Fill gaps between partitions
-    rram_total = _SOC_RRAM_TOTAL[soc]
+    rram_total = get_soc_mem_field(soc, MemField.RRAM_TOTAL, board_dir)
+    rram_base_addr = get_soc_mem_field(soc, MemField.RRAM_BASE_ADDR, board_dir)
```
eivindj-nordic: I'm thinking let's give dts_path as input, then this can be resolved internally in `get_soc_mem_field`. That way we avoid having one variable called board_dir and one called boards_dir, which can be a bit confusing. Especially since the now `boards_dir` variable was previously named `board_dir`. I'd say lets keep that variable's name as it is in main to also not change the input parameters.

PizzaAllTheWay: fixed

doc/_scripts/gen_memory_layouts.py
```diff
@@ -5,7 +5,7 @@
 """Parse bm_* board DTS files and generate SVG memory layout diagrams.
 
 Usage:
-    python gen_memory_layouts.py [--board-dir DIR] [--output-dir DIR]
+    python gen_memory_layouts.py [--boards-dir DIR] [--output-dir DIR]
```
eivindj-nordic: Keep as is.
```suggestion
    python gen_memory_layouts.py [--board-dir DIR] [--output-dir DIR]
```

doc/_scripts/gen_memory_layouts.py
```diff
@@ -133,6 +135,118 @@ def _parse_retained_mem(text: str) -> Partition | None:
...
+def _find_common_dtsi(soc: str, board_dir: Path) -> Path:
+    """Return the path to a SoC's common.dtsi file within board_dir."""
+    matches = sorted(board_dir.glob(f"*_{soc}_cpuapp_common.dtsi"))
+    if not matches:
+        raise FileNotFoundError(f"No *_{soc}_cpuapp_common.dtsi found under {board_dir}")
+    return matches[0]
+
+
+@cache
+def _parse_soc_mem_fields(soc: str, board_dir: Path) -> dict[MemField, int]:
```
eivindj-nordic: And others.
```suggestion
def _parse_soc_mem_fields(soc: str, dts_path: Path) -> dict[MemField, int]:
```

doc/_scripts/gen_memory_layouts.py
PizzaAllTheWay: fixed

doc/_scripts/gen_memory_layouts.py
PizzaAllTheWay: fixed

doc/_scripts/gen_memory_layouts.py
```diff
@@ -363,7 +462,7 @@ def _render_mem_column(mm: MemoryMap, x_offset: float) -> tuple[str, float]:
     fragments.append(
         f'<text x="{bx + _BAR_WIDTH / 2}" y="18" '
         f'text-anchor="middle" font-size="13" font-weight="bold" fill="{_TEXT_TITLE}">'
-        f'{mm.label} ({_fmt_size(mm.total)})</text>'
+        f"{mm.label} ({_fmt_size(mm.total)})</text>"
```
anhmolt: Is this change needed?

anhmolt (dismissed):
Looks good! Changes to the values in the memory maps look sane.

The RRAM size on nrf54lm20 should be 2036KB and have been wrong in this script.

The total size of SRAM now correctly accounts for FLPR and CRACEN reserved memory at the end (0.512KB on nRF54L15, 0.704KB on nRF54LM20 and nRF54LV10).

doc/_scripts/gen_memory_layouts.py
PizzaAllTheWay: fixed

eivindj-nordic (approved):

bjorn-tore-taraldsen:
In the "Memory layout diagram", right after the heading "Select DK → SoC → SoftDevice to view RRAM and SRAM layout diagrams.", I would like to include the following note:

Note: The total SRAM size shown in these diagrams may differ from the total given in the product specification. Reserved regions — "VPR saved context" and "Protected RAM! — are part of the SRAM memory map but are not shown as partitions in the diagrams. For the complete layout, see the Memory map section in the device-specific product specification.

PizzaAllTheWay:
> In the "Memory layout diagram", right after the heading "Select DK → SoC → SoftDevice to view RRAM and SRAM layout diagrams.", I would like to include the following note:
> 
> Note: The total SRAM size shown in these diagrams may differ from the total given in the product specification. Reserved regions — "VPR saved context" and "Protected RAM! — are part of the SRAM memory map but are not shown as partitions in the diagrams. For the complete layout, see the Memory map section in the device-specific product specification.

@bjorn-tore-taraldsen 
Added a note there, please take a look if it needs changing.

doc/nrf-bm/boards/memory_layout.rst
```diff
@@ -82,6 +82,16 @@ Memory layout diagrams
...
 
+.. note::
+   The total SRAM size shown in these diagrams may differ from the total given in the product specification.
+   Depending on your SoC, one or more reserved regions may exist at the high end of the SRAM address range.
+   These are part of the memory map but are not shown as partitions here:
+
+   * **Protected RAM** - Destination address used by the KMU when storing keys, so the CRACEN PSA Crypto driver can access the key material for cryptographic operations. Only present on SoCs with CRACEN. See `Working with the KMU and CRACEN`_.
+   * **VPR saved context** - Used by the VPR (RISC-V coprocessor) to save and restore its context when entering and waking from Hibernate mode. Only relevant if your application uses the VPR. See `Working with the FLPR core`_.
+
+   For the complete SRAM layout, see the :guilabel:`Product overview` → :guilabel:`Memory` → :guilabel:`Memory map` section in the `nRF54L Series product specification`_ for your specific SoC.
```
anhmolt: I think we should keep this shorter, as in the original suggestion by BTT. The explanation of what these reserved regions are used for does not fit here.
```suggestion
.. note::
   The total SRAM size shown in these diagrams may differ from the total given in the product specification.
   Reserved regions, "VPR saved context" and "Protected RAM", are part of the SRAM memory map but are not shown as partitions in the diagrams.
   For the complete layout, see the Memory map section in the device-specific product specification.
```
The added link with more information about the VPR does not mention VPR saved context and is therefore wrong to add here, in my opinion.

The link to the product specification and the instructions on how to navigate to the memory map does not work and I have to guess where to click to get there in the end. The links or page navigation might change as well, so  better to avoid the link here and just direct the reader to the product specification like in the original suggestion by BTT.

PizzaAllTheWay: ok, fixed

doc/nrf-bm/links.txt
```diff
@@ -110,6 +110,7 @@
 .. _`Image tool`: https://docs.mcuboot.com/imgtool.html
 .. _`Vector table requirement`: https://developer.arm.com/documentation/100235/0002/sqf1486724325622
 .. _`Working with the KMU and CRACEN`: https://nrfconnectdocs.nordicsemi.com/ncs/latest/nrf/app_dev/device_guides/kmu_guides/index.html
+
```
anhmolt:
```suggestion
```
stray newline.

anhmolt (approved):

doc/nrf-bm/links.txt
PizzaAllTheWay: fixed

doc/nrf-bm/boards/memory_layout.rst
```diff
@@ -82,6 +82,11 @@ Memory layout diagrams
 
 Select **DK** → **SoC** → **SoftDevice** to view RRAM and SRAM layout diagrams.
 
+.. note::
+   The total SRAM size shown in these diagrams may differ from the total given in the product specification.
+   Reserved regions, "VPR saved context" and "Protected RAM", are part of the SRAM memory map but are not shown as partitions in the diagrams.
+   For the complete layout, see the Memory map section in the device-specific product specification.
+
```
b-gent, line 89:
```suggestion
.. note::
   The total SRAM size in the diagram headers can be smaller than the total SRAM listed at the top of each SoC tab.
   The listed total is the SoC total as described in the Datasheet, while the diagrams cover only the memory mapped by the board devicetree: the KMU reserved area and the ``cpuapp_sram`` node.
   On the nRF54L15, nRF54LM20A, and nRF54LV10A, a reserved region of less than 1 KB at the top of SRAM falls outside the ``cpuapp_sram`` node and is therefore not shown.
   For the complete memory map, including reserved regions, see the documentation for `nRF54L15`_, `nRF54LM20A`_, or `nRF54LV10A`_.

```
This is just a suggestion up for discussion.
I think we should have a clearer message here because currently it looks very confusing that we are stating different values practically next to one another:

Image
Same for other SoCs

doc/nrf-bm/boards/memory_layout.rst
```diff
@@ -252,7 +257,7 @@ Select **DK** → **SoC** → **SoftDevice** to view RRAM and SRAM layout diagra
 
          .. group-tab:: nRF54LV10A
 
-            Total RRAM: 1012 KB | Total SRAM: 192 KB
+            Total RRAM: 1012 KB | Total SRAM: 191 KB
```
b-gent:
```suggestion
            Total RRAM: 1012 KB | Total SRAM: 192 KB
```
This is currently stating that the total SoC SRAM is 191 KB and the diagram shows more than that (191.3). Should be 192 KB for LV10?

Bigger change to seriously consider - since this PR is all about automating the generation of these values to prevent errors, can we get rid of these manually-written hardcoded values from this page?

b-gent (changes requested):

doc/nrf-bm/boards/memory_layout.rst
anhmolt: The size was fixed in https://github.com/nrfconnect/sdk-nrf-bm/pull/834. This PR lags behind.
@PizzaAllTheWay, can you rebase this?

doc/nrf-bm/boards/memory_layout.rst
bjorn-tore-taraldsen: @b-gent Reason for adding the note in the first place was, as you also state, that we have different size number next to each other. 
You are focusing describing what we are covered by our map file, while  initial text focuses on which partitions that we have on the actual RAM that are not part of the partition file (VPR context and Protected RAM). and it is these parts that is the source for the differences in the numbers (and should not be included by our map file)

b-gent: I understand. So the only change I would request is change 'Product Specification' to 'Datasheet' - this is the terminology we are using for nRF54 devices 
@PizzaAllTheWay

doc/nrf-bm/boards/memory_layout.rst
PizzaAllTheWay: Rebased, should be good now

doc/nrf-bm/boards/memory_layout.rst
PizzaAllTheWay: Ok, fixed

bjorn-tore-taraldsen (approved):

b-gent (approved):

PR #831, opened 2026-08-03
doc: cmake: auto gen mem layout on doc build

PizzaAllTheWay:
Automates the generation of the board memory layout SVGs during the doc build. Previously, `doc/_scripts/gen_memory_layouts.py` had to be run manually before building the docs, so the memory layout could go stale. A new `gen-memory-layouts` CMake target now runs it automatically as part of the `nrf-bm` docset build.

It also moves the SVG output from the source tree into the Sphinx build directory, so they're treated as build artifacts and never committed to git. SVGs under `doc/nrf-bm/boards/images/` have been removed accordingly.


doc/CMakeLists.txt
```diff
@@ -303,6 +303,14 @@ add_custom_target(
   COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_LIST_DIR}/versions.json ${HTML_DIR}
 )
 
+add_custom_target(
+  gen-memory-layouts
+  COMMAND
+    ${PYTHON_EXECUTABLE}
+    ${NRF_BM_BASE}/doc/_scripts/gen_memory_layouts.py
```
eivindj-nordic: We should set the path her with e.g. 
set(MEM_LAYOUTS_OUTPUT_DIR ${CMAKE_BINARY_DIR}/nrf-bm/src/boards/images)

```suggestion
    ${NRF_BM_BASE}/doc/_scripts/gen_memory_layouts.py --output-dir ${MEM_LAYOUTS_OUTPUT_DIR}
```

So that this works independently of the build directory used with CMake.

PizzaAllTheWay: fixed

eivindj-nordic (approved):
Looks good! I think the commits can be squashed.

doc/CMakeLists.txt
nordicjm: how can you call something that isn't added until a later commit ?

doc/CMakeLists.txt
```diff
@@ -303,11 +303,12 @@ add_custom_target(
   COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_LIST_DIR}/versions.json ${HTML_DIR}
 )
 
+set(MEM_LAYOUTS_OUTPUT_DIR ${CMAKE_BINARY_DIR}/nrf-bm/src/boards/images)
```
nordicjm: lower case variables names

nordicjm (changes requested):

doc/_scripts/gen_memory_layouts.py
```diff
@@ -11,7 +11,11 @@
 parses every ``.dts`` file that defines flash/SRAM partitions, and writes
 one SVG per configuration into the output directory.
 
-Re-run this script whenever devicetree sources change.
+The doc build (see doc/CMakeLists.txt, target ``gen-memory-layouts``) always
+passes ``--output-dir`` explicitly, pointing at the nrf-bm docset's build
+source directory, so the generated SVGs are build artifacts and are never
+checked into the repository. The default below is only used for manual,
+standalone runs of this script.
```
anhmolt: This should not be here in my opinion. The `gen_memory_layouts.py` script should not have any knowledge on how it is used by any potentially existing CMake scripts.

doc/_scripts/gen_memory_layouts.py
```diff
@@ -23,7 +27,7 @@
 
 NRF_BM_BASE = Path(__file__).resolve().parents[2]
 BOARDS_DIR = NRF_BM_BASE / "boards" / "nordic"
-OUTPUT_DIR = NRF_BM_BASE / "doc" / "nrf-bm" / "boards" / "images"
+OUTPUT_DIR = NRF_BM_BASE / "doc" / "_build" / "nrf-bm" / "src" / "boards" / "images"
```
anhmolt: I don't think we should modify this here. The name of the build folder, here `_build`, is decided when running the doc build with cmake, so it will be wrong to make the assumption here that it is always `_build`.

anhmolt (commented):
Perhaps squash all changes into one commit with the exception of the removal of the `doc/nrf-bm/boards/images/*.svg` files. That can go into a second commit.

doc/_scripts/gen_memory_layouts.py
anhmolt: Can modify the default output directory to the current working directory. That will make more sense when running the script directly without `--output-dir` specified.

```suggestion
OUTPUT_DIR = Path.cwd() / "memory_layouts"
```

doc/CMakeLists.txt
PizzaAllTheWay: fixed

doc/CMakeLists.txt
PizzaAllTheWay: I squashed both commits now.

anhmolt:
> Perhaps squash all changes into one commit with the exception of the removal of the `doc/nrf-bm/boards/images/*.svg` files. That can go into a second commit.

It is also completely fine to remove the images in the same commit.

PizzaAllTheWay:
> Perhaps squash all changes into one commit with the exception of the removal of the `doc/nrf-bm/boards/images/*.svg` files. That can go into a second commit.

Squashed it, kept `doc/nrf-bm/boards/images/*.svg` files in the same commit as well, not separate. .SVG files are part of the same general change.

doc/_scripts/gen_memory_layouts.py
PizzaAllTheWay: fixed

doc/_scripts/gen_memory_layouts.py
PizzaAllTheWay: I took and implemented this suggestion instead https://github.com/nrfconnect/sdk-nrf-bm/pull/831#discussion_r3720222618

PizzaAllTheWay: Ok, fixed

doc/_scripts/gen_memory_layouts.py
```diff
@@ -11,7 +11,9 @@
 parses every ``.dts`` file that defines flash/SRAM partitions, and writes
 one SVG per configuration into the output directory.
 
-Re-run this script whenever devicetree sources change.
+If ``--output-dir`` is not given, the SVGs are written to a default
+location under ``doc/_build/``. The generated SVGs are build artifacts
+and should not be checked into the repository.
```
anhmolt:
```suggestion
If ``--output-dir`` is not specified, the SVGs are written to a folder
``memory_layouts`` in the current working directory.
```

PizzaAllTheWay: fixed

anhmolt (approved):

nordicjm (approved):

bjorn-tore-taraldsen:
I think we should have following NOTE included above the drawing showing the aprtition's 

Note: The total SRAM size shown in these diagrams may differ from the total given in the product specification. Reserved regions—VPR saved context and Protected RAM—are part of the SRAM memory map but are not shown as partitions in the diagrams. For the complete layout, see the Memory map section in the device-specific product specification.

eivindj-nordic:
> I think we should have following NOTE included above the drawing showing the aprtition's
> 
> Note: The total SRAM size shown in these diagrams may differ from the total given in the product specification. Reserved regions—VPR saved context and Protected RAM—are part of the SRAM memory map but are not shown as partitions in the diagrams. For the complete layout, see the Memory map section in the device-specific product specification.

Handled in #820.

doc/nrf-bm/boards/memory_layout.rst
b-gent:
```suggestion
The diagrams are auto-generated from the board devicetree sources during the documentation build, so they always match the current devicetree files.

```

b-gent (changes requested):

doc/nrf-bm/boards/memory_layout.rst
anhmolt: Good catch. This we should update.

PizzaAllTheWay: Good point, fixed

b-gent (approved):

PR #410, opened 2025-10-03
workflow: compliance: Use additional compliance checks

nordicjm:
Adds the compliance checks that are being used in sdk-nrf

PR #833, opened 2026-08-05
treewide: Kconfig and CMake cleanup

eivindj-nordic:
Sort lines in Kconfig and CMake in alphabetic order where the order of the lines does not matter.
Cleanup zephyr_include_directories and zephyr_library_sources.

### DIFF
--- a/drivers/Kconfig
+++ b/drivers/Kconfig
@@ -6,9 +6,9 @@
 menu "Device Drivers"
 
 # zephyr-keep-sorted-start
+rsource "bm_lpuarte/Kconfig"
 rsource "console/Kconfig"
 rsource "flash/Kconfig"
-rsource "bm_lpuarte/Kconfig"
 # zephyr-keep-sorted-stop
 
 endmenu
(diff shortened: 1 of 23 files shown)


anhmolt (approved):

eivindj-nordic:
@nrfconnect/ncs-co-build-system @nrfconnect/ncs-eris Please have a look!

drivers/Kconfig
```diff
@@ -6,9 +6,9 @@
 menu "Device Drivers"
 
 # zephyr-keep-sorted-start
+rsource "bm_lpuarte/Kconfig"
 rsource "console/Kconfig"
 rsource "flash/Kconfig"
-rsource "bm_lpuarte/Kconfig"
```
nordicjm, line 11: how did this even make it in?

nordicjm (approved):

drivers/Kconfig
eivindj-nordic: Good question! At the time this was merged check_compliance workflow was specifying modules to run instead of modules to exclude and this check was not included. It was fixed in https://github.com/nrfconnect/sdk-nrf-bm/pull/410.

michalek-no (approved):

PR #834, opened 2026-08-05
doc: nrf-bm: correct sizes in memory layout page

anhmolt:
The nRF54LM20A has 2036 KB of RRAM, not 2048 KB.
The nRF54LV10A has 192 KB of SRAM, not 191 KB. (Due to the VPR saved context and Protected RAM, not all can be used by application.)


eivindj-nordic (approved):

PizzaAllTheWay (approved):
LGTM

PR #837, opened 2026-08-12
doc: license: update copyright year in documentation and license

anhmolt:
Update year in the copyright footer of the nrf-bm and the Kconfig tabs of the documentation.

### DIFF
--- a/doc/kconfig/conf.py
+++ b/doc/kconfig/conf.py
@@ -18,7 +18,7 @@ NRF_BASE = NRF_BM_BASE / ".." / "nrf"
 # General configuration --------------------------------------------------------
 
 project = "Kconfig reference"
-copyright = "2025, Nordic Semiconductor"
+copyright = "2025-2026, Nordic Semiconductor"
 author = "Nordic Semiconductor"
 # NOTE: use blank space as version to preserve space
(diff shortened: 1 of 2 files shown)


eivindj-nordic (approved):

LICENSE
```diff
@@ -1,7 +1,7 @@
 LicenseID:  LicenseRef-Nordic-5-Clause
 
 ExtractedText: <text>
-Copyright (c) 2025, Nordic Semiconductor ASA
```
eivindj-nordic, line 4: I wonder a bit about this one, in sdk-nrf this file states 2018. https://github.com/nrfconnect/sdk-nrf/blob/main/LICENSE

@carlescufi @b-gent

b-gent: I'd leave the LICENSE at 2025. AFAIK the year there is the year of first publication, which is why sdk-nrf still says 2018. Updating could create an impression that we have somehow modified the license.

b-gent (approved):
Approving , but I recommend to not update the license year.

anhmolt:
> Approving , but I recommend to not update the license year.

Makes sense. Reverted LICENSE file change.

What about the license footer in the docs, would you rather I close this PR?

PR #843, opened 2026-08-19
bluetooth: services: align characteristic add order

PizzaAllTheWay:
test_bm_samples: PR-131

Add characteristics in the same order as their Bluetooth SIG specification tables, so the GATT attribute table matches the documented order inside the specifications.

### DIFF
--- a/doc/nrf-bm/release_notes/release_notes_changelog.rst
+++ b/doc/nrf-bm/release_notes/release_notes_changelog.rst
@@ -136,10 +136,18 @@ Bluetooth LE Services
 
    * Changed :c:member:`ble_scan_filter_data.addr_filter.addr` and :c:member:`ble_scan_filter_data.name_filter.name` to ``const`` in the :c:struct:`ble_scan_filter_data` structure.
 
+* :ref:`lib_ble_service_bms`:
+
+   * Updated the :c:func:`ble_bms_init` function to add the Bond Management Control Point characteristic before the Bond Management Feature characteristic, matching the order in the BMS specification.
+     This changes the GATT attribute handles assigned to these characteristics.
+
 * :ref:`lib_ble_service_dis`:
 
    * Added support for configuring the Device Information Service characteristics at run time through the new :c:struct:`ble_dis_values` structure, passed using the ``values`` field of :c:struct:`ble_dis_config`.
      When ``values`` is ``NULL``, the service is built from the Kconfig defaults as before.
+
+   * Updated the :c:func:`ble_dis_init` function to add the IEEE Regulatory Certification Data List characteristic before the PnP ID characteristic, matching the order in the DIS specification.
+     This changes the GATT attribute handles assigned to these characteristics.
 
 * :ref:`lib_ble_service_mcumgr`:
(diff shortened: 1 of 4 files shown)


eivindj-nordic (approved):

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -136,11 +136,19 @@ Bluetooth LE Services
 
    * Changed :c:member:`ble_scan_filter_data.addr_filter.addr` and :c:member:`ble_scan_filter_data.name_filter.name` to ``const`` in the :c:struct:`ble_scan_filter_data` structure.
 
+* :ref:`lib_ble_service_bms`:
+
+   * Updated the :c:func:`ble_bms_init` function to add the Bond Management Control Point characteristic before the Bond Management Feature characteristic, matching the order in the BMS specification.
+     This changes the GATT attribute handles assigned to these characteristics.
```
b-gent, line 142:
```suggestion
     This changes the GATT attribute handles assigned to these characteristics.
     Clients that cached the attribute handles of this service must rediscover the GATT database after updating the firmware.
```

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -136,11 +136,19 @@ Bluetooth LE Services
...
+
 * :ref:`lib_ble_service_dis`:
 
    * Added support for configuring the Device Information Service characteristics at run time through the new :c:struct:`ble_dis_values` structure, passed using the ``values`` field of :c:struct:`ble_dis_config`.
      When ``values`` is ``NULL``, the service is built from the Kconfig defaults as before.
 
+   * Updated the :c:func:`ble_dis_init` function to add the IEEE Regulatory Certification Data List characteristic before the PnP ID characteristic, matching the order in the DIS specification.
```
b-gent, line 149:
```suggestion
   * Updated the :c:func:`ble_dis_init` function to add the IEEE 11073-20601 Regulatory Certification Data List characteristic before the PnP ID characteristic, matching the order in the DIS specification.
```
this is the name of the charateristic as per [Bluetooth SIG](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/DIS_v1.2/out/en/index-en.html#UUID-11cbfd5c-b7fa-3d12-0f0c-3fdbbd2e4efb) , it is also worth to fix this in ble_dis.rst where we call it "IEEE 11073-20601 Regulatory Certification characteristic".

b-gent (changes requested):

hermabe:
> Add characteristics in the same order as their Bluetooth SIG specification tables, so the GATT attribute table matches the documented order inside the specifications.

Is this a requirement in the profile specifications? Generally the client would do discovery or read by uuid and not rely on any ordering of characteristics.

anhmolt:
> Is this a requirement in the profile specifications? Generally the client would do discovery or read by uuid and not rely on any ordering of characteristics.

You are correct. So I think we should close this. Not necessary to do.

PR #848, opened 2026-08-24
manifest: remove sdk-nrf-bm-internal

eivindj-nordic:
Remove sdk-nrf-bm-internal as it is no longer in use.


eivindj-nordic:
nrf-bm-internal manifest entry expected to be removed, removed DNM.

anhmolt (approved):

lemrey (approved):

PR #856, opened 2026-09-17
lib: bluetooth: peer_manager: write buffer clear and release fixes

anhmolt:
Clear contents of write buffers on release. Release the bonding data write buffer in `auth_status_success_process()` in branches where the content is not to be written to persistent storage.

Note that write buffer allocation leaks have not been an issue because new connections have been able to reuse already allocated buffers for the bonding data. This reuse have been possible because of how temporary peer IDs are allocated based on indices returned with `nrf_sdh_ble_idx_get()`.


lib/bluetooth/peer_manager/modules/peer_database.c
PizzaAllTheWay, line 645:
```suggestion
    write_buffer_record->peer_id = new_peer_id;
	write_buffer_record->data_id = data_id;

	uint32_t nrf_err = write_buf_store(write_buffer_record);

	if ((nrf_err != NRF_SUCCESS) && (nrf_err != NRF_ERROR_RESOURCES)) {
		write_buffer_record_release(write_buffer_record);
	}

	return nrf_err;
```

If the write is queued successfully, BM_ZMS later sends a completion event. This reaches `pdb_pds_evt_handler()`, which releases the buffer.

If `write_buf_store()` fails before queueing, no completion event will arrive, so the buffer would remain allocated. Release it here on terminal errors.

`NRF_ERROR_RESOURCES` is excluded because the buffer is retained for retry. `NRF_ERROR_BUSY` is converted to `NRF_SUCCESS` inside `write_buf_store()` for the same reason.

PizzaAllTheWay, line 600:
```suggestion
	if (buffer_memory == NULL) {
		...
		write_buffer_record_release(write_buffer_record);
		return NRF_ERROR_INTERNAL;
	}
```

`pm_buffer_block_acquire()` has already locked the block at this point. Although `pm_buffer_ptr_get()` returning `NULL` indicates an internal consistency failure, returning directly would leave the record/block reserved. Release it here to prevent a permanent buffer leak.

lib/bluetooth/peer_manager/modules/security_dispatcher.c
PizzaAllTheWay, line 899:
```suggestion
#if defined(CONFIG_SOFTDEVICE_PERIPHERAL)
		aux_sec_params = (role == BLE_GAP_ROLE_PERIPH) ? sec_params : NULL;
#endif /* CONFIG_SOFTDEVICE_PERIPHERAL */

		nrf_err = sd_ble_gap_sec_params_reply(conn_handle, sec_status, aux_sec_params,
						      &sec_keyset);
		if ((nrf_err != NRF_SUCCESS) && (nrf_err != NRF_ERROR_BUSY)) {
			uint16_t temp_peer_id;

			if (pdb_temp_peer_id_get(conn_handle, &temp_peer_id) == NRF_SUCCESS) {
				(void)pdb_write_buf_release(temp_peer_id,
							    PM_PEER_DATA_ID_BONDING);
			}
		}
	}

	return nrf_err;
}
```
When `sec_keyset_fill()` succeeds, it acquires a temporary PDB write buffer and places pointers to that buffer in `sec_keyset`. The buffer must remain allocated while the SoftDevice performs pairing and writes the generated bonding keys into it.

If `sd_ble_gap_sec_params_reply()` returns `NRF_SUCCESS`, the SoftDevice accepted those pointers. Then later `BLE_GAP_EVT_AUTH_STATUS` path in `smd_ble_evt_handler()` will either store the bonding data or release the buffer and clean up for us.

If the call returns a terminal error, the SoftDevice did not accept the keyset. No pairing completion event is guaranteed. Without explicit cleanup, the buffer record remains allocated indefinitely. Release it here using the temporary peer ID.

`NRF_ERROR_BUSY` is excluded because it is transient, Security Manager retries the reply, and the same temporary buffer can be reused.

PizzaAllTheWay, line 820:
```suggestion
		if (nrf_err) {
			LOG_WRN("im_ble_addr_get() returned %s. conn_handle: %d. Link was "
				"likely disconnected.",
				nrf_strerror_get(nrf_err), conn_handle);
			(void)pdb_write_buf_release(temp_peer_id, PM_PEER_DATA_ID_BONDING);
			return NRF_ERROR_INVALID_STATE;
		}
```

Normally, `sec_keyset_fill()` acquires the temporary bonding buffer and `smd_params_reply()` passes its pointers to the SoftDevice. Which will send a  `BLE_GAP_EVT_AUTH_STATUS` event to `smd_ble_evt_handler()`, which will eventually store or release the buffer.

However, in this specific case, `im_ble_addr_get()` fails after the buffer was acquired with error code `NRF_ERROR_INTERNAL`, so `smd_params_reply()` function will skip calling SoftDevice. And because of the error code `NRF_ERROR_INVALID_STATE`, Security Manager does not retry. Therefore, no completion event is guaranteed and the buffer remains reserved. That's why we should release it here.

PizzaAllTheWay (commented):
Good catch on the missing buffer cleanup paths and clearing the buffer contents on release. The overall change looks good.

I left a few suggestions for similar early error paths where a buffer may already be acquired, but no later completion event is guaranteed to release it. Please take a look and let me know whether my understanding of the ownership and retry flow is correct.

PizzaAllTheWay:
I also found a case where the write buffer should be released in the [GATTS Cache Manager](https://github.com/nrfconnect/sdk-nrf-bm/blob/9a8b2286350023bb7668b2457ad8650e437de88c/lib/bluetooth/peer_manager/modules/gatts_cache_manager.c#L184).

```c
				if ((nrf_err != NRF_SUCCESS) &&
				    (nrf_err != NRF_ERROR_NOT_FOUND)) {
					LOG_ERR("pds_peer_data_read() returned %s "
						"for conn_handle: %d",
						nrf_strerror_get(nrf_err),
						conn_handle);
					(void)pdb_write_buf_release(
						peer_id, PM_PEER_DATA_ID_GATT_LOCAL); // Add this here?
					return NRF_ERROR_INTERNAL;
				}
```

`pdb_write_buf_get()` succeeded before this branch, but this early return neither stores nor releases its buffer, while the equivalent exit paths below release it first.

anhmolt:
> I left a few suggestions for similar early error paths where a buffer may already be acquired, ...

Thanks. I will take a closer look at these additional places where write buffer releasing might be needed and do a follow-up PR.

If the current changes looks good then I would like to merge them.

PizzaAllTheWay:
Ok, sounds good

PizzaAllTheWay (approved):

doc/nrf-bm/release_notes/release_notes_changelog.rst
```diff
@@ -143,11 +143,14 @@ Libraries
       * The LESC key agreement handling to clear the static RAM copy of the ECDH shared secret after the secret have been handed over to the SoftDevice using the :c:func:`sd_ble_gap_lesc_dhkey_reply` function.
       * The :kconfig:option:`CONFIG_PM_LESC_GENERATE_NEW_KEYS` Kconfig option to be enabled by default.
         This option forces the use of new ECDH key pair for each pairing procedure.
+      * The write buffer release function in the peer database submodule to erase the contents of the write buffer record on release.
```
anhmolt: temp stuff

lib/bluetooth/peer_manager/modules/peer_database.c
anhmolt: It is not the responsibility of `pdb_write_buf_store()` to release the buffer if it returns an error. That is up to the user.

By design, on success, it will asynchronously release the buffer when the contents have been written to NVM.

anhmolt: The `pm_buffer_ptr_get()` function should not fail here. The successful call to pm_buffer_block_acquire() above ensures this.

lib/bluetooth/peer_manager/modules/security_dispatcher.c
anhmolt: I'll look closer at this.

anhmolt: I'll look closer at this.

anhmolt:
> I also found a case where the write buffer should be released in the [GATTS Cache Manager]

I'll take a closer look.

eivindj-nordic:
@nrfconnect/ncs-bm-doc Please have a look!

eivindj-nordic (approved):

PR #857, opened 2026-09-17
lib: bluetooth: peer_manager: LESC DHkey computation request fixes

anhmolt:
* Treat invalid connection handle as success when replying with a dhkey to the SoftDevice. It signals that the peer disconnected and the DH key is no longer needed.
* Synchronize access to the DHkey computation requested state to avoid some theoretical race conditions between `nrf_ble_lesc_request_handler()` called from main context and `nrf_ble_lesc_on_ble_evt()` called from interrupt context.


PizzaAllTheWay (approved):
LGTM

eivindj-nordic (approved):
Changelog?

anhmolt:
> Changelog?

let us skip that here. Most changes are internal. The first commit (1a30bc29d25a5aeae83f7f41729f8d931653e8d2) touches the API of `nrf_ble_lesc_request_handler()` but the change should not be visible.

